PR #3, opened 2016-02-17
Fixup metapkgs

danielballan:
These are just suggestions. See commit msgs for rationales. There no glaring issues as far as I can see, so good work. I just went through setting up the collection env manually with @hhslepicka today, so the requirements are fresh in my mind.


ericdill:
Awesome thanks!

PR #5, opened 2016-02-19
FIX: Remove xray_vision: misspelled, redundant.

danielballan:


ericdill:
Good catch, thanks. We can blame @tacaswell for this one :smiling_imp:

PR #9, opened 2016-02-23
Add new bluesky, ophyd, databroker versions.

danielballan:


ericdill:
Could you modify the 0.4.0rc2 recipe in-place to become 0.4.0?  I made some substantial edits to the meta.yaml to include a number of dependencies to testing and import *all* the modules to make sure they are all importable.  Doing this found at least one bug in bluesky, so I would prefer to have this minimal level of sanity checking before we deploy tagged builds

ericdill:
https://github.com/NSLS-II/auto-build-tagged-recipes/blob/master/py3/bluesky-0.4.0rc2/meta.yaml

ericdill:
(Note that I am also happy to take over and just make this work if you would rather be doing other stuff)

ericdill:
databroker and ophyd look good, thanks!

ericdill:
Thanks @danielballan ; I took over and did the bluesky modifications I suggested above.  These builds should be available in ~30 minutes

danielballan:
Thanks again.

PR #10, opened 2016-02-23
Remove album from analysis metapkg.

danielballan:
I don't think we need this installed by default. It is *probably* a dead project. Too soon to nuke the album recipe entirely though.

### DIFF
--- a/pyall/analysis-2015_03/meta.yaml
+++ b/pyall/analysis-2015_03/meta.yaml
@@ -19,4 +19,3 @@ requirements:
     - openpyxl
     - scikit-xray >=v0.5.0
     - xray-vision >=v0.3.0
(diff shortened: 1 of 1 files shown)


danielballan:
Where do I got to remove this from the dev analysis env?

ericdill:
I don't think there is a dev analysis env

ericdill:
But :+1: on nuking album. It's spirit will become part of the dashboard concept you are thinking of now.

ericdill:
Did you mean to do this on the analysis 2016_01 metapackage?

https://github.com/NSLS-II/auto-build-tagged-recipes/blob/master/pyall/analysis-2016_01/meta.yaml

Issue #7, opened 2016-02-20
Need a newer version of pims


ericdill:
@danielballan what version of pims can you do `from pims import pipeline`?

ericdill:
I'm getting an error on the csxtools build

ericdill:
```
===== testing package: csxtools-0.1.2-np110py34_0 =====
import: 'csxtools'
/opt/conda/envs/_test/lib/python3.4/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
/opt/conda/envs/_test/lib/python3.4/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
Traceback (most recent call last):
  File "/opt/conda/conda-bld/test-tmp_dir/run_test.py", line 25, in <module>
    import csxtools
  File "/opt/conda/envs/_test/lib/python3.4/site-packages/csxtools/__init__.py", line 8, in <module>
    from .utils import get_fastccd_images
  File "/opt/conda/envs/_test/lib/python3.4/site-packages/csxtools/utils.py", line 4, in <module>
    from pims import pipeline
ImportError: cannot import name 'pipeline'
TESTS FAILED: csxtools-0.1.2-np110py34_0
```

ericdill:
no rush, im done for the evening and will check again tomorrow, thanks!

danielballan:
Latest tag 0.3.3 should work.

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Fri, Feb 19, 2016 at 4:24 PM -0800, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


===== testing package: csxtools-0.1.2-np110py34_0 ===== import: 'csxtools' /opt/conda/envs/_test/lib/python3.4/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
 warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.') /opt/conda/envs/_test/lib/python3.4/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take
[...]

PR #14, opened 2016-03-02
BUG: Stop building ncurses for now

ericdill:
@tacaswell reports that build #7 (the current build in
auto-build-tagged-recipes is causing segfaults.  He will work on fixing
this, but the solution for now is to use build #6.


ericdill:
I have removed the build 7 package from nsls2-tag and am going to self-merge so that this build that is causing segfaults does not re-appear.  Sorry for the hassle.

PR #15, opened 2016-03-02
MNT: Re-introduce configure flag to ncurses

ericdill:
I'm bumping the build number so that this *should* automatically replace any builds with build_number = 7, which apparently are segfaulting!

cc @tacaswell


ericdill:
Note that this reverts the change that I incorporated from @tacaswell 's enh-more PR on conda-prescriptions.  

https://github.com/NSLS-II/conda-prescriptions/pull/106/files#diff-0f105a1e2d17a1d9b0e8ee308dd96d98L6

ericdill:
the ncurses build passed on circle-ci. Merging so this is available by 2:30 internally

PR #16, opened 2016-03-03
Update bluesky recipe for 0.4.3

ericdill:
I know this looks a little crazy, but this is an accurate reflection of the bluesky dependencies...  Maybe we should consider splitting out some of these modules and subpackages into their own repositories?

attn @danielballan


danielballan:
I think we all agree that hardware_checklist.py should be moved out, probably related to https://github.com/NSLS-II/wishlist/issues/101. Maybe the suspenders as well.

danielballan:
For today I was just hoping for `s/0.4.2/0.4.3/`. This is not a great time to do something so disruptive to the recipe.

ericdill:
I am merely trying to prevent things like https://github.com/NSLS-II/bluesky/issues/371

ericdill:
I will take 

>I think we all agree that hardware_checklist.py should be moved out, probably related to NSLS-II/wishlist#101. Maybe the suspenders as well.

to mean that I can ignore deps that are only in hardware_checklist and suspenders

ericdill:
I will do as you wish for now, but I think we are being a little fast and loose with the dependencies here.  I am only going to add `xray-vision` as a dep to make sure that NSLS-II/bluesky#371 does not occur again.  I'm also going to sort the dependencies (other than python) and the imports, so that it is easier to quickly glance at the modules and dependencies

danielballan:
If conda had a way of expressing "optional" deps, that is what I would advocate for. I think in the future that it will. In the meantime, I want it to be possible to install bluesky without databroker/xray-vision/pyepics etc. That's where I'm coming from. We used to solve this issue by distributing a working `collection` metapackage.

ericdill:
> We used to solve this issue by distributing a working collection metapackage.

Whoever installed the collection environment at SRX failed to install the collection metapackage.  Is it currently broken?

ericdill:
updated previous comment with an additional question

py3/bluesky-0.4.3/meta.yaml
```diff
@@ -40,9 +40,13 @@ test:
     - bluesky
     - bluesky.broker_callbacks
     - bluesky.broker_examples
+    - bluesky.callbacks
+    - bluesky.callbacks.core
+    - bluesky.callbacks.olog
```
tacaswell: I suspect that this will fail if you do not have olog configure properly.  If it does not find the configuration file it is looking for it will _on import_ use raw_input to ask for information.  This is why we do not have any import tests in the pyOlog recipe.

tacaswell: nm, it has always been there...no idea how that is working.

ericdill: It's working because you only import pyolog inside of the callback
https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/olog.py#L148

PR #17, opened 2016-03-04
ENH: add hdf5-lz4

tacaswell:
This is needed to read eiger/dectris hdf5 files.

### DIFF
--- a/nonpy/hdf5-lza-0.2/meta.yaml
+++ b/nonpy/hdf5-lza-0.2/meta.yaml
@@ -1,0 +1,21 @@
+package:
+    name: hdf5-lz4
+    version: 0.2
+
+source:
+    git_url: https://github.com/dectris/HDF5Plugin.git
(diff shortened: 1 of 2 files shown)


nonpy/hdf5-lza-0.2/meta.yaml
```diff
@@ -0,0 +1,21 @@
+package:
+    name: hdf5-lz4
+    version: 0.2
+
+source:
+    git_url: https://github.com/dectris/HDF5Plugin.git
+    git_tag: 8fa813aa369c15127c551cfdeebe4cbc090b4687  # no tags in repo
+
+build:
+    number: 3
```
ericdill: If this is a new recipe this should be 0

tacaswell:
Not new, there is already a version of this on lightsource2 and on the hand-built nsls2 channel internally

ericdill:
Cool, thanks!

PR #19, opened 2016-05-02
Add metadatastore v0.4.1

danielballan:
Note: we never made one for v0.4.0, tagged six days ago, but there's no practical reason to go back for it. v0.4.1 has important fixes.

### DIFF
--- a/pyall/metadatastore-0.4.1/build.sh
+++ b/pyall/metadatastore-0.4.1/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+sed -i 's|__CONDA_BUILD_PLACEHOLDER__|'$PREFIX'/etc|' metadatastore/conf.py
+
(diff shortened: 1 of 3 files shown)


ericdill:
It is easier to maintain a pile of recipes if we only keep the ones that we actively care about.  Do we need metadatastore 0.3.0 or 0.3.2 still?

danielballan:
For MDS, it might be worth keeping those around for another cycle just in case. (But I'm less worried about old versions of bluesky, databroker, doct...)

danielballan:
Actually: I forgot for a second that they will always be in the git history of this repo. So, sure.

ericdill:
Alright. If possible, just edit the recipes in-place. Makes it easier to maintain hundreds of these recipes

ericdill:
I'll deal with the old recipes. You have enough other stuff on your plate, but moving forward, could you edit recipes in place?

danielballan:
Sure thing.

PR #21, opened 2016-05-07
BLD: Added h5py and hdf5 for SWMR

stuwilkins:
Attn @ericdill 
cc: @tacaswell @danielballan 

This adds h5py and hdf5 libraries that work with SWMR. I have built and they work for me ok. Could you please have a look and if good merge, I need these packages on CSX. Thanks!


ericdill:
lgtm. thanks @stuwilkins

PR #22, opened 2016-05-09
ffmpeg is not building on the controls network

ericdill:
Getting the following error for ffmpeg:

'fatal: unable to connect to source.ffmpeg.org:\n'
 'source.ffmpeg.org: Name or service not known\n'


ericdill:
@stuwilkins any idea if it is possible to access `git://source.ffmpeg.org/ffmpeg.git` from the controls network?  `git clone git://source.ffmpeg.org/ffmpeg.git` fails behind the nsls2 firewall

stuwilkins:
HI @ericdill I don't think that the git protocol will make it through the firewalls. Can you use http access for ffmpeg? 

What about using [this mirror](https://github.com/FFmpeg/FFmpeg) and then using:

https://github.com/FFmpeg/FFmpeg.git

ericdill:
Now there's a good idea! I'll try that and report back, thanks!!

ericdill:
That fixes it. Thanks @stuwilkins

PR #24, opened 2016-06-03
Sort run requirements. Add suitcase

ericdill:
@danielballan @tacaswell @klauer Do you want to change any of the other
requirements for the analysis metapackage for the 2nd cycle?  Here I am doing 
two things: 

(1) sorting the packages alphabetically so it is easier to check
    what is included in the meta.yaml
(2) Adding suitcase to the analysis environment.


pyall/analysis-2016_02/meta.yaml
```diff
@@ -1,23 +1,25 @@
+{% set version = "2016_02" %}
```
tacaswell: I really like this pattern

pyall/analysis-2016_02/meta.yaml
```diff
@@ -1,23 +1,25 @@
...
+    - bluesky  # [py3k]
     - channelarchiver
+    - databroker
+    - datamuxer
+    - ipywidgets
+    - jupyter
     - lmfit ==0.8.3
     - matplotlib
     - openpyxl
+    - pymongo
```
tacaswell: probably do not need this explicitly (as it is dep of mds/fs)

ericdill: IIRC we were pinning it previously.  Now we *might* not need to pin?  Or maybe we do need to pin at `pymongo >=3.2`?  If we don't need to pin, I agree this can go away

danielballan: I think we might need to pin. Would it be better to do that in the mds recipe?

ericdill: > Would it be better to do that in the mds recipe?

Yeah.  Pinning in the library recipe where it actually matters is almost certainly a good policy

ericdill: bump @tacaswell

tacaswell:
:+1: but it needs a rebase.

ericdill:
Thanks @danielballan :+1:

pyall/analysis-2016_02/meta.yaml
tacaswell: I suggest removing this and making sure it is in the mds/fs recipies

tacaswell: Nevermind :sheep:

PR #26, opened 2016-06-08
ENH: add mongo configs for SRX/SMI

tacaswell:


ericdill:
We should change these to be env vars that get set/unset on activate/deactivate, but this is fine for now

tacaswell:
We use the files so that jupyter hub (which does not strictly activate it's env) can use these.

I suspect that the best solution is, as @cowanml has suggested, a service that you query with BL + other service, and it returns a valid configuration dictionary.  I also strongly suspect that this has a name and there are a bucket of existing tools that just do this.

arkilic:
Checkout conftrak in LIX tools. Hugo and I already discussed this and writing the first iteration this week

PR #28, opened 2016-06-09
ENH: Use env vars for dev configuration

ericdill:
attn @tacaswell Any reason why this is a bad idea?  I keep getting issues
where the config cannot seem to load values out of the specified 
configuration file location and it greatly annoys me. This fixes that by 
just using env vars. 

Also note the addition of the pre-unlink.sh script that will remove those
two files if/when this package is removed. We should be more careful about
keeping track of the files that we add with our conda packages and clean 
up after ourselves when those packages are removed.

### DIFF
--- a/nonpy/config-dev/build.sh
+++ b/nonpy/config-dev/build.sh
@@ -1,28 +1,48 @@
 #!/bin/bash
 
-mkdir -p $PREFIX/etc
-ETC=$PREFIX/etc
-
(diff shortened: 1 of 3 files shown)


tacaswell:
@danielballan do we ever want to use the dev configuration with jupyter hub?  I am thinking not.

danielballan:
Given that all of this will go away when we move away from singleton MDS/FS objects, whatever makes @ericdill's life easiest is fine by me.

ericdill:
Thanks for weighing in!

PR #29, opened 2016-06-13
Regenerate config packages at build time and then build them

ericdill:


ericdill:
Is there any objection to automatically generating the package content on circle-ci from the "config.yaml" file and then building that automatically generated package and uploading it?

ericdill:
cc @tacaswell

tacaswell:
What is the source of the failures here?

ericdill:
1. an api change in how conda build handles the GIT_* environmental variables. This affects the ophyd-2.x conda package. Do we still need this?
2. A bug/feature in the way that `build_from_yaml` executes code.  It looks at the build_directive.yaml and finds folders of recipe folders to build and then builds a big list.  This is actually a problem because I am using sed to replace the local directory with the one from github. tl;dr I can fix it, it's not a problem.

klauer:
Agreed - ophyd is py3k only now for many reasons.

PR #32, opened 2016-07-25
Update

licode:
waiting for ophyd's new version.


licode:
ready to go. @tacaswell ?

tacaswell:
I think these recipes will fail to build due to sha failures?

licode:
updated.

tacaswell:
Did you mean to commit those changes to the scripts?

licode:
I think this should be ready to go. I tested the build process which works fine.

PR #34, opened 2016-07-26
DEV: update on meta files

licode:
for beamline upgrade.


tacaswell:
can you also move over the minimum version for FS on databroker? 

After that self-merge this.

PR #36, opened 2016-07-29
Add event-model tagged recipe.

danielballan:
We were missing an event-model recipe in recipes-tag. This adds one for the newly-tagged v1.1.0.


danielballan:
attn @arkilic @licode

Issue #38, opened 2016-08-17
remove boltons from metadatastore (in all the places)


tacaswell:

licode:
Only remove boltons from metadatastore, or all over the places, like filestore, databroker?

tacaswell:
Remove it from anyplace we do not actually need it.

PR #40, opened 2016-08-29
MNT: boltons are removed

licode:
Done.

### DIFF
--- a/recipes-dev/bluesky/meta.yaml
+++ b/recipes-dev/bluesky/meta.yaml
@@ -31,6 +31,7 @@ requirements:
     - historydict
     - tifffile
     - event-model
+    - boltons
 
 test:
   requires:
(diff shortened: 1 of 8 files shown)


tacaswell:
I think bluesky still uses boltons.

licode:
OK, I thought you wanted it to be deleted from any places.

licode:
Added it back.

tacaswell:
Only on the projects that do not actually have it as a dependency any more.

You can quickly check if a project uses it by `git grep boltons` in the source tree.

licode:
Right, that's what I am doing.

arkilic:
databroker also uses boltons [here](https://github.com/NSLS-II/databroker/blob/0365af849a3ef5fa0e8d254bff72c3c3a1da2f37/databroker/core.py#L585)

licode:
Filestore also needs that. So only remove boltons from metadatastore.

licode:
ready to go? anyone?

arkilic:
Closes #38

PR #44, opened 2016-09-08
DEV: meta file to dev folder

licode:
The recipes in -tag folder can be shipped to conda-forege again. We will keep -dev.


danielballan:
Great. Merge at will.

arkilic:
Thank you @licode !

licode:
Sure. It is better to send recipes to both -tag and -dev later.

PR #48, opened 2016-09-29
Add metapackages for collection, analysis.

danielballan:
This PR adds metapackages, similar to conda's `environment.yml` idea but with slightly different feature set. We tried this once before and abandoned it. Maybe it will stick this time.

* The `analysis` metapackage contains most of our stuff, including bluesky because the callbacks are useful on saved data.
* The `collection` metapackage depends on the `analysis` one, and it adds ophyd and pyolog.
* The version scheme is not semantic versioning: it is the date plus a micro version for corrections (e.g., `2016-09-29.0`.

### DIFF
--- a/metapackages/analysis/meta.yaml
+++ b/metapackages/analysis/meta.yaml
@@ -1,0 +1,24 @@
+{% set version = "2016-09-29.0" %}
+
+package:
+  name: analysis
+  version: {{ version }}
+
+requirements:
+  run:
+    - python==3.5.1
+    - ipython==5.1
+    - ipywidgets==5.2.2
+    - jupyter
+    - bluesky==0.6.4
+    - filestore==0.6.0
+    - suitcase==0.5.1
+    - databroker==0.6.2
+    - historydict==1.2.0
+    - doct==1.0.3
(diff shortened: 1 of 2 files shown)


tacaswell:
Maybe a terser date? `16Q4.0` ?

danielballan:
That was my impulse (that's what we did last time) but yesterday you said, "The correct way to do this is to specify the whole date."

tacaswell:
I need to get my memory checked apparently.

danielballan:
:- D

For real, I can see the case either way. A "quarter" is prettier; a date is in some ways "easier" because it's completely clear.

danielballan:
Feel free to push to my branch and merge.

ericdill:
These recipes also exist in recipes-tag. [collection] (https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/collection/meta.yaml) and [analysis] (https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/analysis/meta.yaml)

Probably want to remove them from there to avoid future confusion

PR #49, opened 2016-09-29
ENH: add hxnfly recipe

tacaswell:


tacaswell:
don't merge just yet, have a uncommited change that needs to be pushed

PR #52, opened 2016-10-10
MNT:Add missing mds requirement to mdservice conda recipes

arkilic:
Attn @licode


licode:
Looks good to me.

arkilic:
Shld this update conda installation as soon as merged or you need to update the script?

licode:
Installation with nsls2-dev will be updated immediately. But you need to put a new version in recipe in order to make nsls2-tag update.

arkilic:
👍

PR #59, opened 2017-01-04
DEV: version update for collection and analysis

licode:


tacaswell:
`YYQCC.N` would be terser, but we may not want to mess with the existing scheme so they continue to sort in the correct order.

licode:
Can we use build number 0, 1, 2 for different versions?
build:
  number: 0

tacaswell:
different builds of the same version should be semantically interchangeable, the minor version bumps will not be.

licode:
good to know. thanks.

PR #61, opened 2017-01-05
DEV: add doct back

licode:


tacaswell:
Does this also need a build.sh?

licode:
only meta file from conda forge

PR #63, opened 2017-01-05
DEV: doct recipes, tested

licode:


licode:
already tested, merged.

PR #73, opened 2017-01-09
hxnfly

licode:
Still keep version 17Q1.0, as update is done at hxn. We will update version later.

### DIFF
--- a/recipes-tag/03-id-hxn-collection/meta.yaml
+++ b/recipes-tag/03-id-hxn-collection/meta.yaml
@@ -11,6 +11,7 @@ build:
 
 requirements:
   run:
+    - hxnfly ==0.0.2
     - hxntools
     - tqdm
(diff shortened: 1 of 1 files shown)


recipes-tag/03-id-hxn-collection/meta.yaml
```diff
@@ -7,10 +7,11 @@ package:
   version: {{ year }}{{ cycle }}.{{ version }}
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   run:
+    - hxnfly ==0.0.2
```
danielballan: I suggest changing this to `>=0.0.2` so that we don't need remember to update this when `hxnfly` updates.

licode: good point

PR #76, opened 2017-01-09
Update chxtools recipe target to v0.0.3

arkilic:


danielballan:
cc @yugangzhang: Let us know if the latest chxtools picked up any new dependencies since v0.0.2. I am merging this for now so we can deploy it and test it out.

PR #77, opened 2017-01-09
Remove setuptools from pysmbc

arkilic:


arkilic:
@licode did this recipe work?

danielballan:
I gather it won't work until [this file](https://github.com/NSLS-II/debian-with-miniconda/blob/master/Dockerfile) is updated to include the lib and the builder is relaunched from that image.

arkilic:
https://github.com/NSLS-II/debian-with-miniconda/pull/1

PR #80, opened 2017-01-10
enh: add xpd packages

tacaswell:


danielballan:
I merged since you sound like you're in a hurry over there, but strictly speaking the xpdacq recipe should include a bluesky dependency.

PR #105, opened 2017-03-17
ENH: add caRepeater to epics-base

tacaswell:


cowanml:
cainfo is handy for debugging while you're at it...
http://epics.web.psi.ch/style/training/handouts/e_basehandouts/node11.html

stuartcampbell:
@cowanml - good suggestion, is there anything else we should have ? Feel free to open a PR.  :)

PR #109, opened 2017-03-23
channel archiver added

licode:
per @tacaswell  request.


danielballan:
Do we want to package that NSLS-II fork instead? It is "8 commits ahead, 9 commits behind" the upstream branch. :- \ One of the fixes on our branch is py3 compat.

danielballan:
Bump @licode @tacaswell

tacaswell:
Definitely build from our version.

danielballan:
This issue is blocking databroker + 3.6. It should be high-priority to support external collaboration here: https://github.com/NSLS-II/databroker/issues/140

licode:
OK, I will change this to our nsls2 version.

danielballan:
Notice that the latest commit on channelarchiver master includes py3-compat fixes and that the latest tag is several commits behind. Also, it almost looks like that tag is a typo (they go 0.0.4, 0.0.5, 0.0.5.6!). Maybe we should tag channelarchiver master as 0.0.6 and target this recipe at that.

licode:
lets close this if there is no dependency on archiver?

danielballan:
Sorry for the runaround, @licode.

PR #110, opened 2017-03-29
Fix pims

danielballan:


danielballan:
FYI, @licode, we just needed to build the required dependencies tifffile and slicerator for 3.6. No special error here. I think the error messages you pasted into Slack may have just *seemed* confusing because they were incomplete.

PR #113, opened 2017-03-29
readline

tacaswell:


tacaswell:
Self-merging to turn the crank and try to unblock other things.

PR #122, opened 2017-04-05
DEV: run packages building without docker

licode:


licode:
simple script, merged.

PR #126, opened 2017-04-14
ENH: update attrs

CJ-Wright:


recipes-tag/attrs/meta.yaml
```diff
@@ -22,16 +22,13 @@ requirements:
   run:
     - python
     - zope
+    - zope.interface
```
tacaswell: do we really need these things at runtime?!

It looks like these are only testing or helper dependencies, see https://github.com/python-attrs/attrs/blob/master/setup.py

PR #136, opened 2017-05-05
different folder for condaforge recipes

licode:


licode:
anything else from cf? @tacaswell

licode:
pyfai, slicerator

danielballan:
This should wait for the next cycle now -- probably want to avoid creating chaos at this point.

tacaswell:
Why do pyfai, slicerator, pympler get completely deleted?

licode:
closed. not implemented soon.

PR #138, opened 2017-05-09
DEV: new envs for 2017Q2

licode:
any package missing here?


danielballan:
Tom had mentioned wanting to include https://ipyvolume.readthedocs.io/en/latest/ in the analysis stack this cycle.

licode:
sure, will add that

PR #166, opened 2017-05-25
after bug fix

licode:


licode:
https://github.com/NSLS-II-HXN/hxntools/pull/9

PR #170, opened 2017-06-07
BUG: Specify pyolog version

stuartcampbell:
Seems like there are issues with 4.2.0 so we will fix 4.1.0 for now.


stuartcampbell:
Closing as we'll just change the version of the pyolog package in the '-tag' channel.

PR #189, opened 2017-08-18
hdf5, h5py, lz4-c

licode:
h5py based on hdf5  1.10.1.


licode:
@tacaswell

recipes-tag/h5py/bld.bat
```diff
@@ -0,0 +1,5 @@
+"%PYTHON%" setup.py configure --hdf5="%LIBRARY_PREFIX%"  --hdf5-version=1.8.18
```
tacaswell: Even though we are not building windows, this should be corrected.

recipes-tag/h5py/run_test.py
```diff
@@ -0,0 +1,24 @@
...
+import h5py.h5g
+import h5py.h5i
+import h5py.h5l
+import h5py.h5o
+import h5py.h5p
+import h5py.h5r
+import h5py.h5s
+import h5py.h5t
+import h5py.h5z
+import h5py.utils
```
tacaswell: Can you add a test in here that it is finding the correct version of hdf5?

tacaswell: `h5py.h5.HDF5_VERSION_COMPILED_AGAINST`

licode: I installed latest version 2.7 of h5py from conda-forge, and tried this command,
```
In [3]: h5py.h5.HDF5_VERSION_COMPILED_AGAINST
   ...:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-34c104899b6f> in <module>()
----> 1 h5py.h5.HDF5_VERSION_COMPILED_AGAINST

AttributeError: module 'h5py.h5' has no attribute 'HDF5_VERSION_COMPILED_AGAINST'
```

tacaswell: Ah, that is only on master, try ` h5py.version.hdf5_version_tuple`

licode: done

PR #196, opened 2017-09-07
BLD: update lmfit version

chiahaoliu:
old ``lmfit``  uses deprecated ipython widget hence cause error importing

### DIFF
--- a/recipes-tag/scikit-beam/meta.yaml
+++ b/recipes-tag/scikit-beam/meta.yaml
@@ -30,7 +30,7 @@ requirements:
         - six
         - xraylib    # [not win]
         - scikit-image
-        - lmfit 0.8.3
+        - lmfit 0.9.7
 
 test:
     imports:


licode:
This will break lots of functions in skbeam. We plan to update lmfit maybe in next cycle, not currently. There are also some speed issues in new version of lmfit.

chiahaoliu:
so is there any suggested action regarding this dependency when we have to use ``ipython`` ?

licode:
can you try   
```
conda install ipywidgets
```

danielballan:
@chiahaoliu Did Li's suggestion resolve your issue?

chiahaoliu:
Yes, this fix works for my local system (Mac) but I haven't had a chance to test on the workstation at beamline. May I test this on Monday and report the status?

danielballan:
That sounds great. Thanks.

chiahaoliu:
The ``ipywidgets`` dependency does the job. Close issue for now.

Issue #205, opened 2017-09-20
Add oscars conda package


danielballan:
I've been manually installing `oscars==1.37.1.dev3` using pip for @dhidas.

We should make a conda package so it can be added to our 'analysis' conda metapackages and installed automatically everywhere.

stuartcampbell:
Cool - @licode can you have a look at it ?

licode:
Sure, I will add this.

dhidas:
Hi,

Thanks.  I'm happy to do this with you guys.  If we can have a discussion at some point that might help me.  There are some complications with conda, namely compiler name mangling that differs from nvidia's nvcc which I haven't been able to reconcile so far.  If there are no GPU machines the solution is simple, which is just compile without it, but if there are it's worth it (a factor of > 1000 in many cases) to get this working with conda.  Of course, I can upload the manylinux1 binaries to pypi in any case...  but I'd like to know really the best course of action.

Also, I'm planning a major upgrade in the next ~1.5 weeks, maybe even earlier...

Cheers,

-Dean

dhidas:
Hi,

Is there any chance we could quickly upgrade to the latest version: pip install oscars==2.0.2  ?

I don't know if it's a lot of effort or not.  Ideally it would be good to have it up in a week if that's not asking too much?

And if there is a way for me to make it easier for everyone in the future I'm all about that...

Cheers,

-Dean

stuartcampbell:
@licode Have you made any progress?

licode:
working on this. I find there are lots of c code in the github source code, and also calculations related to gpu and multithreading. This may be a bit challenging, and I will keep everyone updated.

danielballan:
@dhidas Could you summarize for @licode what you told me a couple weeks ago about the dependency on a newer version of certain libraries than one finds on typical conda build systems (~CentOS 5)?

dhidas:
The GPU tools need to be compiled with the nvcc compiler which is only available for cent os 6 or greater (and similar for other distros), so it won't exactly fit the manylinux1 standard, but perhaps would work well enough...or should I not go there?  I haven't found nvcc in any conda package yet, but they do distribute the libs in cudatoolkit.

To compile with the gpu I would normally:
make
python setup.py install

...however there are some issue when using conda, namely that nvcc name mangling was different from the conda gcc in some cases.  One could use nvcc to compile and link everything, but I didn't get that to play nice with setuptools, though I imagine it is possible.

[...]

PR #208, opened 2017-10-12
add oscars

licode:
local build runs ok. https://github.com/NSLS-II/lightsource2-recipes/issues/205#issuecomment-335983026

### DIFF
--- a/recipes-tag/oscars/meta.yaml
+++ b/recipes-tag/oscars/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "oscars" %}
+{% set version = "2.0.2" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "b159fba768604ed6d00afdef9cdfb4ce570a0c11bce7dcbee43556aecd189ba6" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.{{ file_ext }}'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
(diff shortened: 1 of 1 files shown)


licode:
@danielballan

licode:
@dhidas , can you also check if any dependency is missing.

dhidas:
This seems ok.  mpi4py is not required so you can leave it out you like.  Unless someone has a networked cluster I believe the internal multi-threadding is better anyways.

If you're compiling it you'll need a c++11 compliant compiler, but maybe you are pulling the manylinux1 binaries?  These at the moment are the same (for linux only).  In the future when I figure out how to package the binary for a pseudo-manylinux the binary will contain libraries which can only be made with nvcc, which I guess is difficult with conda at the moment...

licode:
I think I will leave mpi4pi there just in case it is used. I follow setup.py where c++11 compiler is defined for linux. Let's give it a try, and see if it works ok.

licode:
The conda package oscars will be shown at https://anaconda.org/lightsource2-tag/ once it is ready, so you can try
```
conda install oscars -c lightsource2-tag
```

or you can try the following if at beamline machine
```
conda install oscars -c nsls2-tag
```

dhidas:
Great!  And a question.. It will show up on notebook.nsns2 ?

licode:
We need to re-install it first. But you can test it on your local machine first when the package is ready.

licode:
The package is ready to use https://anaconda.org/lightsource2-tag/oscars

stuartcampbell:
Awesome @licode

dhidas:
Tested and it does work!

Also, does this automatically pick up the master branch, or release, or something from github?  I notice that my v2.0.2 is already there...

Many thanks, this is really great.

tacaswell:
The tag to use is set https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/oscars/meta.yaml#L2

When you do a new release just open a PR adjusting that number.

PR #210, opened 2017-10-27
Bugfix version update

dhidas:
Fix for importing arbitrary function as efield.

### DIFF
--- a/recipes-tag/oscars/meta.yaml
+++ b/recipes-tag/oscars/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "oscars" %}
-{% set version = "2.0.2" %}
+{% set version = "2.0.4" %}
 {% set file_ext = "tar.gz" %}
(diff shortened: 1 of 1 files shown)


licode:
Thanks for the update. But I will close this, and further simply this recipe. Because the current recipe still needs sha value. I will send a PR to review.

licode:
@dhidas  Your github shows the release the number is 2.00.04. https://github.com/dhidas/OSCARS/tags
That's the number used to build, instead of 2.04. Is this ok? 

We can build 2.02 previously because the recipe searched source from pip which has 2.02. But now we want it to search github release.

dhidas:
It seems better to keep the numbering the same.  I tagged 2.0.4 in github and will keep with this scheme without zfill.  Seem fine to you?

licode:
ok, I will update the version to 2.0.4.

PR #211, opened 2017-10-27
DEV: simplified version for oscars, depending on github release

licode:
@dhidas  next time we only need to update version number based on the new recipe.


licode:
version number updated, ready to go.

licode:
2.0.4 should be ready to use.

PR #216, opened 2017-11-07
DEV: new version of bluesky and ophyd

licode:
1.0.0 version after this.

### DIFF
--- a/recipes-tag/bluesky/meta.yaml
+++ b/recipes-tag/bluesky/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.10.3" %}
+{% set version = "0.11.0" %}
 
 package:
     name: bluesky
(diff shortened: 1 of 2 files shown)


danielballan:
Checking that the dependencies haven't changed:

ophyd:

```
$ git diff v0.7.0 v0.8.0 requirements.txt
✔
```

bluesky:

```
$ git diff v0.10.2 v0.11.0 requirements.txt
diff --git a/requirements.txt b/requirements.txt
index fbb4752..408f717 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ cycler
 event-model
 historydict
 jsonschema
-matplotlib
 numpy
 super_state_machine
 toolz
✔
```

...and of course we want to keep the matplotlib dependency in the conda recipe, because we make the conda recipes err on the side of inclusiveness of optional deps.

PR #217, opened 2017-11-08
v1.0.0 for bluesky and ophyd

licode:
1.0.0 is ready to build as previous versions are done according to https://github.com/NSLS-II/lightsource2-recipes/pull/216


licode:
@danielballan

danielballan:
Thanks for taking care of this!

danielballan:
attn @teddyrendahl This merge should kick off a build machine that uploads to lightsource2-tag every hour or so.

PR #219, opened 2017-12-04
OSCARS version update to 2.0.6

dhidas:
Only OSCARS version change.  Nothing else.


licode:
sorry to get back late on this. 2.0.6 should be ready to use.

PR #222, opened 2017-12-07
CLN: Remove environments, which we store in ansible playbook now.

danielballan:


mrakitin:
Is it safe to delete those envs?

danielballan:
Yes, there are copies of these same files in NSLS-II/playbook (a private repo that DAMA members have access to) and it is more useful to keep them there.  Better not to maintain them in two places, where they might get out of sync.

mrakitin:
Thanks for the explanation, Dan!

PR #223, opened 2017-12-07
Draft update for 18Q1.

danielballan:
There may be additional revisions before we deploy but this
is good to merge.

- Bump the version number.
- Add @dhidas 'oscars' package to the standard 'analysis' env.
- Also add 'dask' and 'distributed'.


mrakitin:
Looks good to me.

PR #226, opened 2017-12-20
FIX: correct HDF5_PLUGIN_PATH to use $CONDA_PREFIX

mrakitin:
Based on conversation in Slack.


danielballan:
Context for future reference: it seems like conda changed the name of this env var under us. Have not confirmed that, but this works.

jrmlhermitte:
:+1:

mrakitin:
Thanks!

tacaswell:
I am just going to delete the current package from both the public and private channels (or it won't build!)

PR #227, opened 2017-12-21
Updated chxtools v2.0.1

mrakitin:
Updated chxtools v2.0.1. There is a lot of work to do, but it's a good step towards usability and reproducibility.

### DIFF
--- a/recipes-tag/chxtools/meta.yaml
+++ b/recipes-tag/chxtools/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.0.4" %}
+{% set version = "2.0.0" %}
 
 package:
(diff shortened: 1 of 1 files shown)


recipes-tag/chxtools/meta.yaml
```diff
@@ -33,5 +33,5 @@ test:
     - chxtools.pims_readers
 
 about:
-  home: https://github.com/NSLS-II/dataportal
```
danielballan: Good catch.

tacaswell:
https://github.com/NSLS-II-CHX/chxtools/releases/tag/v2.0.1 for reference

PR #229, opened 2017-12-26
Bump isstools

jrmlhermitte:
bump isstools

### DIFF
--- a/recipes-tag/isstools/meta.yaml
+++ b/recipes-tag/isstools/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.3" %}
+{% set version = "1.3" %}
 
 package:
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
Oh I forgot, there are two [https://github.com/NSLS-II-ISS/isstools/blob/v1.3/requirements.txt](requirements) for it: `numexpr` and `pysmbc`. Does that need to be added here?

licode:
I think numexpr needs to be added. I git grep pysmbc, but didn't find it in use. A simple way to test is  building the package locally to see if it runs ok.

jrmlhermitte:
thanks. it should be `git grep smbc`. We definitely tested and found we needed it. i will add thanks!

PR #231, opened 2017-12-28
DEV: mirror packages continuously

licode:


scripts/mirror.py
```diff
@@ -305,7 +305,7 @@ def cli():
                 message = message + ('\n\nCheck the log at {} on {}'
                                      ''.format(args.log, os.uname()))
                 slack_api.chat.post_message(slack_channel, message)
-            sys.exit(1)
+            #sys.exit(1)
```
tacaswell: Why did you take out the sys.exit ?

licode: mirroring will stop if upload of some pkg fails. I want to make sure mirroring process can continue to run. We can always check which one failed from log file.

PR #235, opened 2018-01-08
BUG: lower case for pychx

licode:
@tacaswell and @mrakitin


mrakitin:
Should I change the original package https://github.com/NSLS-II-CHX/pyCHX?

licode:
No. Only need to change the name shown in package building.

PR #238, opened 2018-01-10
Updated dependencies

jrmlhermitte:
DO NOT MERGE 
works built, loaded on chx workstation, but want to finish the whole profile update process first

### DIFF
--- a/recipes-tag/chxtools/meta.yaml
+++ b/recipes-tag/chxtools/meta.yaml
@@ -18,19 +18,28 @@ requirements:
     - setuptools
 
   run:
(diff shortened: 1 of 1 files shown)


licode:
There may be a conflict between pyqt5 and enaml. I can double check.

jrmlhermitte:
the dependencies work, however, one extra step was necessary:
`pip install pyqt5==5.7.1`

mrakitin:
Note: we should not have conda-installed PyQt5 package to avoid a problem with GLIBC-2.14 is not installed.

PR #239, opened 2018-01-11
CHX: fix dependencies for collection-2018-1.0

mrakitin:


licode:
thanks for updating this. @danielballan recommended before to use playbooks to store all the beamline collection instead of here, maybe let's double check with him on this?

jrmlhermitte:
I tried to build and I get:
Error: bad character '\' in package/version: 2018\-1.0

@licode is there a way to keep dashes?

tacaswell:
I suspect not, as conda uses `-` for sematic separation.  I would go with `.` or `C`

jrmlhermitte:
Ok. I have two PR's to make for two different extra packages. Can we agree that "c" is the convention used?
So `2018C1.0`? thanks!

mrakitin:
Changed dash to underscore until we decide what's a better symbol ('C' or '_').

jrmlhermitte:
Can we vote on this and decide?
:+1: - `_`
:-1: - `C`

mrakitin:
OK, updated to `C` :)

jrmlhermitte:
looks good, also built again just in case..... so looks like it's "C" merging! :-)

mrakitin:
Great, thanks for reviewing and merging!

PR #242, opened 2018-01-12
eiger-io: remove filestore dependency

mrakitin:


tacaswell:
Can you also bump the build number on both?

mrakitin:
OK. Should I also tag a new version of the packages?
- https://github.com/NSLS-II-CHX/eiger-io/tags
- https://github.com/NSLS-II-CHX/chxtools/tags

mrakitin:
@tacaswell?

jrmlhermitte:
If I understand correctly, the build number forces a build? (is this true?)

Since we have a patch that may come in from [here](https://github.com/NSLS-II-CHX/eiger-io/pull/13)
should we hold off for one day and update the version then?

We could also tag chxtools with a later version number while we're at it since it's slightly behind master and we're going to test `eiger-io` at chx.

mrakitin:
Done with version bumps. Also, updated the URL after transferring of the ownership of eiger-io to [NSLS-II](https://github.com/NSLS-II/eiger-io).

mrakitin:
Also updated url/fn to git_url/git_rev for pychx.

licode:
LGTM

mrakitin:
Thanks for merging it, @licode!

PR #245, opened 2018-01-17
Apply new naming scheme.

danielballan:
Note that the _environments_ installed this cycle have already been manually
updated to use the new naming scheme. This change makes the names of the metapackages
consistent. It will intentionally cause the builder to build a 2018-1.0 version
that is identical to the 18Q1.0 version.

### DIFF
--- a/recipes-tag/collection/meta.yaml
+++ b/recipes-tag/collection/meta.yaml
@@ -4,7 +4,7 @@
 
 package:
   name: collection
-  version: {{ year }}-{{ cycle }}.{{ version }}
+  version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
   number: 2
--- a/recipes-tag/collection/meta.yaml
+++ b/recipes-tag/collection/meta.yaml
@@ -1,10 +1,10 @@
-{% set year = "18" %}
-{% set cycle = "Q1" %}
+{% set year = "2018" %}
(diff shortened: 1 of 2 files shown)


recipes-tag/collection/meta.yaml
```diff
@@ -4,7 +4,7 @@
 
 package:
   name: collection
-  version: {{ year }}-{{ cycle }}.{{ version }}
+  version: {{ year }}C{{ cycle }}.{{ version }}
```
stuartcampbell: Is is "C" or _ better ?  Discuss ? ;-)

tacaswell: Can we also drop the '20' as we are not time travelers and hopefully this will get overhauled before 2100? ;-)

danielballan: We renamed the _environments_ from 18Q1.0 to 2018-1.0 for consistency with the directory naming scheme. The only issue here in the dash, which is allowed in environment names but disallowed in version specs.

mrakitin: `2018_1.0` seems more consistent with `2018-1.0`.

mrakitin: ... similarly to `eiger-io` package is imported as `eiger_io`. But I don't mind if we use `C`.

PR #249, opened 2018-01-17
scikit-beam version bump

jrmlhermitte:
version bump


jrmlhermitte:
good idea! done

mrakitin:
Thanks!

PR #254, opened 2018-01-19
Xpd fix

CJ-Wright:
@jrmlhermitte @tacaswell I think this is ready to go in

### DIFF
--- a/recipes-tag/28-id-2-xpd-analysis/meta.yaml
+++ b/recipes-tag/28-id-2-xpd-analysis/meta.yaml
@@ -4,7 +4,7 @@
 
 package:
   name: 28-id-2-xpd-analysis
-  version: {{ year }}-{{ cycle }}.{{ version }}
+  version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
   number: 0
--- a/recipes-tag/28-id-2-xpd-analysis/meta.yaml
+++ b/recipes-tag/28-id-2-xpd-analysis/meta.yaml
@@ -1,0 +1,15 @@
+{% set year = "2018" %}
+{% set cycle = "1" %}
+{% set version ="0" %}
+
+package:
+  name: 28-id-2-xpd-analysis
+  version: {{ year }}-{{ cycle }}.{{ version }}
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - streamz
+    - shed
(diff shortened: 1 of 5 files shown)


recipes-tag/28-id-2-xpd-analysis/meta.yaml
```diff
@@ -0,0 +1,16 @@
+{% set year = "2018" %}
+{% set cycle = "1" %}
+{% set version ="0" %}
+
+package:
+  name: 28-id-2-xpd-analysis
+  version: {{ year }}-{{ cycle }}.{{ version }}
```
danielballan: We have learned that conda won't allow `-` in a version specification, so we changed `-` to `C`.

CJ-Wright:
@danielballan conda-forge is where I got all the recipes from
I used the following xonsh script
```python
import os
from contextlib import contextmanager
@contextmanager
def indir(d):
    """Context manager for temporarily entering into a directory."""
    old_d = os.getcwd()
    ![cd @(d)]
    yield
    ![cd @(old_d)]
a = ['xpdsim', 'xpdtools', 'shed', 'xpdacq', 'xpdan']
t = 'https://raw.githubusercontent.com/conda-forge/{}-feedstock/master/recipe/meta.yaml'
for b in a:
    os.makedirs(b, exists_ok=True)
    with indir(b):
        if 'meta.yaml' in g`*`:
            rm meta.yaml
        wget @(t.format(b))
```
Credit to @scopatz for the `indir` function which is vendored from rever. This script could be expanded to include more recipes and automated with Travis-CI and `doctr` to auto update all the recipes if there was interest.

CJ-Wright:
Although that script could be made more robust. We could have wget pull the entire folder and only replace data if it gets something.

danielballan:
Looks good. I'll let @jrmlhermitte merge.

jrmlhermitte:
looks good to me. thanks for all your efforts in this CJ, in maintaining conda-forge (making this also easier). I like how you have the sha256's in. is that automated?

danielballan:
I second @jrmlhermitte's thanks!!

CJ-Wright:
Of course, I'm working with @justcalamari on automating a bunch of conda-forge version bumping, which could make this easier too especially if we have the auto pull from cf code.
We rely heavily on DAMA code so it is very important that things get packaged ASAP and often.

PR #256, opened 2018-01-19
Auto cf

### LINKED ISSUE #255: Auto pull recipes from conda forge
It might be nice to have a system which trolls through conda forge for new versions of the recipes.

CJ-Wright:
Closes #255 
@jrmlhermitte @danielballan 
I think this should work, it worked on my local. Once everyone is happy with this they should:
1. turn the `totally_not_a_travis.yaml` file into a valid `.travis.yaml`
1. Run doctr to get the keys setup properly
1. Figure out which branch to push to so we don't end up with recursive things (one may consider separating the branches/repos which hold the recipes for general PRing and the branches/repos which hold the recipes for the internal server to read/build) (one could also consider putting this script in a separate repo, cloning this repo, updating everything, and then pushing back to this repo.
1. Turn on travis

### DIFF
--- a/scripts/cf_recipe_update.xsh
+++ b/scripts/cf_recipe_update.xsh
@@ -16,7 +16,7 @@ for d in ['../recipes-dev', '../recipes-tag']:
         t = 'https://raw.githubusercontent.com/conda-forge/{}-feedstock/' \
             'master/recipe/meta.yaml'
         for b in a:
-            os.makedirs(b, exists_ok=True)
+            os.makedirs(b, exist_ok=True)
             with indir(b):
                 url = t.format(b)
                 request = requests.get(url)
--- a/scripts/cf_recipe_update.xsh
+++ b/scripts/cf_recipe_update.xsh
@@ -10,7 +10,7 @@ def indir(d):
     ![cd @(d)]
     yield
     ![cd @(old_d)]
-for d in ['recipes-dev', 'recipes-tag']:
+for d in ['../recipes-dev', '../recipes-tag']:
     with indir(d):
         a = g`*`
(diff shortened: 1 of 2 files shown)


tacaswell:
Can this be done with plain python or plain bash?

CJ-Wright:
Probably, PRs welcome.

danielballan:
We have no xonsh users on staff, so we are not really able to maintain this as is. As you speak xonsh, would you be willing to translate it for us?

CJ-Wright:
I won't be able to translate this right now. I may be able to later in the week. If there is some interest in learning xonsh here are the docs: 
- http://xon.sh/ 
- http://xon.sh/tutorial.html 

Xonsh is fairly intuitive, the deps are very light, and the devs are very responsive.

danielballan:
I really love xonsh; I hope it takes over the world; I agree with @tacaswell that this is not a good enough reason to add yet-another-thing that some onboarding member of staff would have to learn in order to maintain our already-complex build stack.

Sorry for the thanks-for-the-milk-can-we-have-a-cookie nature of this request, but some help translating this would be great when you have the time!

CJ-Wright:
@mrakitin this is the PR with auto conda-forge updating.

mrakitin:
@CJ-Wright, thanks, but I'm not sure how we can use it with our current building approach. Can you explain?

CJ-Wright:
This would issue PRs into lightsource2-recipes with updates from the conda-forge recipes which are in lightsource2. Those PRs could be merged and then building would continue as normal. The main benefit of this is that new versions/updates to recipes from the community would be PRed over to lightsource2-recipes automatically.

A potential extension of this would be to build a graph and check that all deps are on lightsource2-recipes or defaults, but that could be done later.

PR #260, opened 2018-01-23
Added py4xs

jrmlhermitte:
Lin Yang has requested we add py4xs.

> Just want to mention that we use py4xs (it's on pypi but not on github) for data processing and it would be nice to have it next time when you update the conda environment.

It's not on github, so we have to use pypi


danielballan:
> It's not on github, so we have to use pypi

That's more than fine. PyPI is probably a better canonical source, in fact.

PR #263, opened 2018-01-25
Added eiger-io and analysis metapackage

### LINKED ISSUE #181: Need an ISR analysis metapackage that includes eiger-io
This manually patched in their current conda env.

jrmlhermitte:
Closes #181 

(we missed this somehow)


jrmlhermitte:
@danielballan i was not sure what you meant by analysis but i added it to analysis and collection anyway (since both witll need it)

jrmlhermitte:
thanks @mrakitin ! :-)

danielballan:
:+1: 

Agreed both need it. Sometimes I just make the analysis metapackage a dependency of the collection -- since analysis-time dependencies are typically also collection-time dependencies but not vice versa. I'm not enamored with that approach, though, because inevitably we'll find some example where something is needed in an analysis env but never in a collection env. (Jupyter-related tools might be one such example.) And one less layer of indirection is probably a good thing.

PR #264, opened 2018-01-25
Fixed analysis metapackage

jrmlhermitte:
fix to analysis metapackage
@danielballan

### DIFF
--- a/recipes-tag/04-id-isr-analysis/meta.yaml
+++ b/recipes-tag/04-id-isr-analysis/meta.yaml
@@ -3,7 +3,7 @@
 {% set version ="0" %}
 
 package:
-  name: 04-id-isr-collection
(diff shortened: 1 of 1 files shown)


danielballan:
Did the original merge supercede the actual 04-id-isr-collection? Better check that we don't have a packaged named 04-id-isr-collection that is actually 04-id-isr-analysis.

danielballan:
I see. Well, the way to do it would be to delete the bad file using the web UI, and then increase the build number the recipe by 1 so that the auto-builder re-processes it.

jrmlhermitte:
thanks. i dont think it's necessary this time since the meta.yaml is the same. the folder name is all that was different and that should have no consequence on the package. correct?

thanks for this info though, it is very useful

danielballan:
Agreed.

PR #269, opened 2018-02-06
ENH: add hdf5-lz4 as a dependency for eiger-io

mrakitin:


danielballan:
I don't think eiger-io is usable without hdf5-lz4, but would like a sanity check from @jrmlhermitte on this, given his CHX experience.

danielballan:
Strictly speaking, Eiger might not always compress with LZ4 but in practice at NSLS-II it always does, and this is mostly a facility-specific package at this point, so we are take the pragmatic approach.

If you discover this PR someday and have a reason to want it out, ask!

danielballan:
Shoot -- forgot to bump that build number. Can you submit a new PR, @mrakitin?

mrakitin:
Done, see #270.

PR #271, opened 2018-02-15
Copy only linux part from conda-forge devel

licode:
https://github.com/conda-forge/matplotlib-feedstock/tree/devel/recipe

@tacaswell

### DIFF
--- a/recipes-dev/matplotlib/build.sh
+++ b/recipes-dev/matplotlib/build.sh
@@ -1,0 +1,727 @@
+
+
+
+
+
+
+<!DOCTYPE html>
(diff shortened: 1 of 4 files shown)


danielballan:
This looks like the entire HTML page from GitHub that displays the code for humans, not the code itself. Example: https://github.com/NSLS-II/lightsource2-recipes/pull/271/files#diff-c7e9b2de2467b3a7a8de6a69e94d2e9e

Perhaps a misaimed `wget`? Use the GitHub's raw view (or, clone).

licode:
good catch.

PR #272, opened 2018-02-17
Pull in protobuf from conda-forge

stuartcampbell:
This is required as a dependency of the new mysql connector.


stuartcampbell:
I'm going to be bad an self-merge! 😟

PR #275, opened 2018-02-20
up to micro-version 20

dhidas:
correct python error.  minor.


licode:
2.0.18 is also ready.

PR #281, opened 2018-03-14
update historydict (fixed a bug)

jrmlhermitte:
@danielballan


jrmlhermitte:
thanks again @tacaswell for finding this

licode:
recommended to use new format without sha256 for later, i.e., https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/bluesky/meta.yaml

jrmlhermitte:
Ok thanks!

PR #283, opened 2018-03-27
tomopy from conda-forge

licode:
@mrakitin


mrakitin:
Great, thanks very much, @licode!

PR #291, opened 2018-04-16
add xpd stack updates

CJ-Wright:


licode:
we need to put this at recipes-tag folder.

mrakitin:
It will be tagged via a separate PR in a few days once it's tested at PDF.

licode:
If a pkg is built based on a given version number, we need to put the recipe at recipe-tag folder. recipe-dev folder is used to build pkg on each update of master branch. You can compare the bluesky folder under -tag and -dev.

PR #293, opened 2018-04-23
rc's for PDF with configs

CJ-Wright:


CJ-Wright:
@licode would it be possible for me to get a merge? I'd like to install this out at PDF today.

licode:
Last week we updated our server which is used to build packages. We need more time to config everything correct in the server to make sure package building run successfully. I am not sure if that can be done today, but will keep you updated.

CJ-Wright:
Ok, currently I'm working with source. But if this works I'd like to get back onto the gold standard.

licode:
Around 40-50 mins. It also depends on which pkg you are building.

PR #297, opened 2018-04-25
auto remove in docker

licode:


scripts/nsls2-dev-build.sh
```diff
@@ -20,7 +20,7 @@ docker pull $IMAGE_NAME
 
 echo "Running the docker container"
 
-cat << EOF | docker run -i \
+cat << EOF | docker run -i --rm\
```
mrakitin: I would add an extra blank space just in case. Same for the below.

licode: good catch.

mrakitin:
Looks great to me, thanks @licode!

licode:
thanks a lot!

PR #298, opened 2018-04-26
CHX: updated analysis extra package version

mrakitin:
Per CHX user's complaint.


mrakitin:
Thanks @licode and @jrmlhermitte!

PR #303, opened 2018-04-30
Update hkl requirements

jrmlhermitte:
hkl requires gsl==2.2.1 as it looks for a specific library file (libgsl.so.19).
We can recompile hkl to use the newer version of gsl (2.4) later. But for now, it is a good idea to keep this dependency, or else we'll see an error of the form (as conda will install the latest version by default):
```
Error: g-invoke-error-quark: Could not locate hkl_factories: libgsl.so.19: cannot open shared object file: No such file or directory 
```

### DIFF
--- a/recipes-tag/hkl/meta.yaml
+++ b/recipes-tag/hkl/meta.yaml
@@ -17,7 +17,7 @@ requirements:
         - autoconf
         - automake
         - gobject-introspection
-        - gsl
+        - gsl==2.2.1
         - libtool
         - m4
         - pkg-config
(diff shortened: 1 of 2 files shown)


mrakitin:
Just a general question, which package are we talking about? If I search for https://pypi.org/project/gsl/, it's a package with a single version 0.0.1. In https://anaconda.org/lightsource2-tag we don't have such a package.

mrakitin:
Got it, never mind, sorry for the noise :).
https://www.gnu.org/software/gsl/

jrmlhermitte:
yeah, they're C libaries we supply with conda so we don't have to worry about installing them at the system level.
I added some more strict dependencies. I found that it worked with a specific version of pygobject3 we have compiled but not the other. Better to be strict than not. It did not work in py3.5 either. We will have to revisit later.

mrakitin:
looks good to me.

jrmlhermitte:
thanks @mrakitin ! :-)

PR #304, opened 2018-05-02
updated hkl version

jrmlhermitte:
We're now at v0.3.12.

This could also be considered v0.4. We change minor number when a new feature is added that is backwards compatible. Since the limits never really worked, this could also be considered a feature.
I'll keep as v0.3.12 (so just change the patch number) but can change ti v0.4 if someone has a different opinion.


jrmlhermitte:
thanks @tacaswell !

PR #305, opened 2018-05-02
OSCARS minor version change

dhidas:
Now has time-dependent E&B input fields.  RK4 step splitting for numerical stability in high E-fields.


mrakitin:
Thanks for the update @dhidas! Looks good to me.

PR #308, opened 2018-05-05
Lightflow

jrmlhermitte:
Lightflow has quite a few deps. Let's start it off on qas and iss only first.

dep tree installed (pushed [here](https://anaconda.org/jrmlhermitte/repo) first)
1. lightflow
    1. pymongo
    2. celery
        1. case
        2. kombu
            1. amqp
                1. vine
        3. billiard
    3. redis-py
    4. ruamel.yaml
    5. colorlog


recipes-tag/pymongo/meta.yaml
```diff
@@ -0,0 +1,47 @@
+{% set version = "3.6.1" %}
```
tacaswell: we don't need pymongo, it is the deaults channel .

tacaswell: unless we need better than 3.4.0

recipes-tag/redis-py/meta.yaml
```diff
@@ -0,0 +1,49 @@
+{% set name = "redis-py" %}
```
tacaswell: We already have this on the anaconda channel.

recipes-tag/pymongo/meta.yaml
jrmlhermitte: unfortunately, it's set by the developer [here](https://github.com/AustralianSynchrotron/lightflow/blob/master/setup.py#L47).
We need pymongo 3.6.0

I can ask to see if he can downgrade this requirement if adding this is a problem. It's very likely he doesn't need 3.6.

tacaswell: fair enough.

recipes-tag/redis-py/meta.yaml
jrmlhermitte: not sure how i missed it (might have forgotten `-c anaconda` when manually building for lightflow the first time around). just pushed the changes.
I also searched for all these packages on the anaconda channel : `anaconda.org/anaconda` and did not find others. good catch!

PR #310, opened 2018-05-09
Strict version on hkl for csx

jrmlhermitte:
We should enforce hkl>=0.3.12 from now on for CSX (and other beamlines eventually)

### DIFF
--- a/recipes-tag/23-id-1-csx1-collection/meta.yaml
+++ b/recipes-tag/23-id-1-csx1-collection/meta.yaml
@@ -1,5 +1,5 @@
 {% set year = "2018" %}
-{% set cycle = "1" %}
+{% set cycle = "2" %}
 {% set version ="0" %}
 
 package:
@@ -7,7 +7,7 @@ package:
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 1
+  number: 0
 
 requirements:
   run:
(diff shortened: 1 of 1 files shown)


recipes-tag/23-id-1-csx1-collection/meta.yaml
```diff
@@ -1,10 +1,10 @@
-{% set year = "17" %}
-{% set cycle = "Q1" %}
+{% set year = "2018" %}
+{% set cycle = "1" %}
```
mrakitin: Should it be `2` already?

recipes-tag/23-id-1-csx1-collection/meta.yaml
```diff
@@ -1,10 +1,10 @@
...
 {% set version ="0" %}
 
 package:
   name: 23-id-1-csx1-collection
-  version: {{ year }}{{ cycle }}.{{ version }}
+  version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
   number: 1
```
mrakitin: `number: 0`?

recipes-tag/23-id-1-csx1-collection/meta.yaml
jrmlhermitte: thanks! updated

recipes-tag/23-id-1-csx1-collection/meta.yaml
jrmlhermitte: thanks! updated

PR #313, opened 2018-05-09
pyzbar wrapper of zbar lib

mrakitin:
Should be merged along with #311.

Both zbar and pyzbar packages were successfully built and tested locally on https://hub.docker.com/r/nsls2/debian-with-miniconda/.

### DIFF
--- a/recipes-tag/pyzbar/run_test.py
+++ b/recipes-tag/pyzbar/run_test.py
@@ -1,0 +1,3 @@
+import pyzbar
+import pyzbar.pyzbar
+from pyzbar.pyzbar import decode
--- a/recipes-tag/pyzbar/build.sh
+++ b/recipes-tag/pyzbar/build.sh
@@ -1,0 +1,1 @@
+sudo apt-get install libzbar0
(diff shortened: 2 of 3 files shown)


recipes-tag/pyzbar/run_test.py
```diff
@@ -0,0 +1,3 @@
+import pyzbar
+import pyzbar.pyzbar
+from pyzbar.pyzbar import decode
```
jrmlhermitte: I'm not familiar. Will this get automatically tested? Also why not just put it in the test import key of the meta.yml?

mrakitin: Yes, it's automatically tested. The reason is that you cannot directly import it with `import pyzbar.pyzbar.decode`.

jrmlhermitte:
Looks good to me. merged, thanks @mrakitin !

mrakitin:
Thanks @jrmlhermitte!

recipes-tag/pyzbar/run_test.py
mrakitin: I haven't found an example how you can include `from pyzbar.pyzbar import decode` into the `meta.yml` directly. If you encountered something like this, I'll be happy to get rid of the `run_test.py` file.

mrakitin: Some docs about test files: https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html?highlight=run_test#test-section

PR #314, opened 2018-05-10
Chx serial

jrmlhermitte:
CHX needs pyserial to talk to a pressure pump


jrmlhermitte:
closing to avoid confusion. Yes, we need one or the other. if people vote against the other (pyserial for all beamlines), then we need this one. thanks!

mrakitin:
Thanks for the clarification!

PR #316, opened 2018-05-10
add pyserial to collection env

jrmlhermitte:
I think this is a useful library for the beamline as a whole. CHX needed it yesterday (#314) but I think we could maybe just make this a facility wide dep.

This allows scientists to quickly integrate serial devices into their beamlines, which can be a useful thing to do once in a while.

If this is merged, don't merge #314


danielballan:
This seems like a sensible facility-wide dep to me. Once up a time, I would have said, "No, everything should go through EPICS." But now that we have caproto, these are no longer mutually exclusive.

PR #318, opened 2018-05-10
remove strict dep

jrmlhermitte:
hkl py fix: should not be string python dep


mrakitin:
I like this update for future releases of Python.

PR #322, opened 2018-05-11
bump eiger-io

jrmlhermitte:
bump eiger-io from addition of a `frame_num` datum kwarg for a different type of read [here](https://github.com/NSLS-II/eiger-io/pull/14).


mrakitin:
LGTM

PR #326, opened 2018-05-14
BLD: update to h5py 2.8.0

tacaswell:
This will let us use numpy 1.14 on the floor this cycle


tacaswell:
This is slightly different that the conda-forge recipe as 1.12 is the oldest version of numpy that the anaconda channel has numpy built for 3.6.

PR #327, opened 2018-05-14
Add kafka

jrmlhermitte:
To get kafka to work, I had to upgrade a few libs, including openssl. The openssl one is problematic, as it means we're no longer using the one from anaconda.
I suggest we try it anyway, and downgrade if it causes problems at beamlines, and open an issue to use anaconda's version once they upgrade to the newest version.


jrmlhermitte:
do not merge: i may have found a simpler solution.

jrmlhermitte:
changed, no longer requires openssl .

good to go

PR #335, opened 2018-05-15
Atom: update version and URL

mrakitin:
DO NOT MERGE YET

Lets wait until a newer version of atom is released: https://github.com/nucleic/atom/releases, so that the fix for Python 3.6 (https://github.com/nucleic/atom/pull/79) is in there.

TODO: increase the version number.


mrakitin:
This package now exists in the `defaults` channel, and the corresponding recipe in this repo was already archived, so no action is needed. Closing.

PR #336, opened 2018-05-15
Fix pysmbc package

mrakitin:
This solves the issue with building of the pysmbc package.

**Limitations:**
- I had to remove the `test` section, since it required the `smbclient` library to be installed in advance. We either need to build it separately, or assume it's already installed on a target machine.

### DIFF
--- a/recipes-tag/pysmbc/meta.yaml
+++ b/recipes-tag/pysmbc/meta.yaml
@@ -1,14 +1,17 @@
+{% set name = "pysmbc" %}
 {% set version = "1.0.15.8" %}
+
 package:
-  name: pysmbc
+  name: {{ name|lower }}
   version: {{ version }}
+
 source:
-    url: https://pypi.io/packages/source/p/pysmbc/pysmbc-{{ version }}.tar.bz2
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.bz2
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
interesting, so the "#" caused issues for build as well? good to know!

mrakitin:
No, it was a strange indent in front of `url:`, looks like it was a tab symbol or some invisible UTF symbol.

mrakitin:
Thanks for merging, btw!

mrakitin:
Another problem was that the `setup.py` is not a standard one, which didn't accept `--single-version-externally-managed --record=record.txt`, that's why I removed that.

jrmlhermitte:
ah, I understand what happenened. I hit tab to add the url and it gave 4 spaces instead of two.
So previously, my first commit [here](https://github.com/NSLS-II/lightsource2-recipes/commit/62c2c83afc80f4a4ea8025259671d5a49ad67cd6) was fine but the second one where I quickly changed the url had an extra two spaces [here](https://github.com/NSLS-II/lightsource2-recipes/commit/210516e6c9d2c62f4e5949740421384bd5997487). I didn't test the second one. I just tested the url.
I think we definitely need to add a linter or something to review packages.

I asked because i had added that comment to remove that flag. I could not run with that flag and wasn't sure why, but i guess now we understand it's because it's setuptools specific.

good to know we can trace the issue!

PR #338, opened 2018-05-15
Xraylarch

jrmlhermitte:
Should we put a recipe for larch as well? Here is what I got working. I added a separate version for lmfit, but can rather absorb it into lmfit if not.

### DIFF
--- a/recipes-tag/asteval/meta.yaml
+++ b/recipes-tag/asteval/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "asteval" %}
+{% set version = "0.9.12" %}
+{% set sha256 = "38f3b0592cae7e7f65adc687e37aad1824a8e518245603a29ec33258277e779b" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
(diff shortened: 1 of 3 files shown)


mrakitin:
The [built file](https://anaconda.org/lightsource2-tag/xraylarch/files) is 2.7 kB only. The reason is below:
```
++ unset _CONDA_PYTHON
++ unset CONDA_PATH
++ [[ -n 4.2.37(1)-release ]]
++ hash -r
+ python setup.py install --single-version-externally-managed --record=record.txt
Checking dependencies....

=== Cannot Install Larch, these REQUIRED Modules are missing:
 numpy, scipy, matplotlib, h5py, sqlalchemy, requests, six, psutil, peakutils, asteval, lmfit, termcolor

 Please read INSTALL for further information.
```
Unfortunately, last release 0.9.37 (March 26) didn't include my PR https://github.com/xraypy/xraylarch/pull/240 (Apr 5), where I took care or installing the requirements before the installation of xraylarch.

https://github.com/NSLS-II/lightsource2-recipes/pull/341 should take care of the old installer.

jrmlhermitte:
awesome! Now i understand why I could build locally. Definitely better practice to try installing requirements. Thanks for the fix and explanation!

mrakitin:
Welcome! I struggled enough at ISS and with the xraylarch PR, hopefully they will release a newer version soon, which will be more pip-compatible.

PR #348, opened 2018-05-17
MNT: add nsls-ii as collection dependency

tacaswell:
discovered at XFP


mrakitin:
Oh, do we also need it in analysis?

PR #351, opened 2018-05-19
ENH: improve building workflow

mrakitin:
## Summary:
I propose a new building approach -- one package per docker container.

### Pros:
- One package per docker container, meaning complete isolation of building process of each package.
- Each docker container is named now (the format is `<package_name>-<%Y%m%d%H%M%S>`), which means we can better monitor what's building at any given moment by `docker ps -a`.
- Avoid a long operation of bulk testing of the builds/versions of _all_ the packages (around 130!) before actual build starts -- it takes about 30 minutes with the old approach. This is achieved by the configurable period of time (2 days by default) to look back in git log that allows us to build just those packages which were recently updated.

### Cons:
- Possible overhead related to recreation of the whole conda environment (takes ~1 minute per package), so if all the packages are requested to be built, it may take long(er) time.
- If used in a cron job, the job may still repeat the operations, which previous iterations are still doing, so we need to be careful with setting the periodicity of those iterations. Nevertheless, this problem exists even with the old approach.

### Tests:
- Successfully tested locally with docker containers on `12-id-smi-collection` package (multiple packages are built and uploaded/mirrored).
- Successfully tested on freyja under my account (a package was built and uploaded/mirrored).
- Benchmarks:
  - building 11 packages took 24m39.619s
  - building 1 package took 2m18.076s

### Post-actions:
- Remove `12-id-smi-collection`, builds 1-4 from https://conda.nsls2.bnl.gov/nsls2-tag/12-id-smi-collection/files (I don't have permissions to remove packages):
image

### Possible improvements:
- Make the order of the packages random, so cron iterations may cover more packages quicker.
- Make the selection procedure smarter, but it's hard to know before the packages/dependencies list is build in docker, so it seems we can't do it trivially. This may require to maintain a local database with the information about the built packages (in databroker? and run builds with bluesky plans? 😃).
- Run a few package builds in parallel/in chunks.
- Other ideas?

### DIFF
--- a/recipes-tag/12-id-smi-collection/meta.yaml
+++ b/recipes-tag/12-id-smi-collection/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   run:
--- a/recipes-tag/12-id-smi-collection/meta.yaml
+++ b/recipes-tag/12-id-smi-collection/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 4
+  number: 0
 
 requirements:
   run:
--- a/recipes-tag/12-id-smi-collection/meta.yaml
+++ b/recipes-tag/12-id-smi-collection/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 1
+  number: 3
 
 requirements:
   run:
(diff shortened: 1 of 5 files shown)


tacaswell:
In `build.py` there is logic to walk the `meta.yaml` files and sort out what the build order needs to be.  I suspect that could either be pulled up to spawn dockers in the right order or passed in as a list to the docker (so back to one docker per-run, but the biggest time saving is going to be the 'only rebuild the new stuff').

May be worth putting a file with a timestamp in XDG_RUNTIME_DIR (or /tmp) with the time it was last run   and use that as the time horizon?  The first time it gets run we have to pay the full cost, but after that it should be fast.

mrakitin:
@tacaswell's suggestions can be addressed in the later revisions of this workflow. Self-merging for now to speed things up on the floor.

Issue #349, opened 2018-05-18
Conflict of scikit-beam and xraylarch in packages


mrakitin:
scikit-beam requires lmfit=0.8.3, but xraylarch needs lmfit>=0.9.9. That should be resolved on scikit-beam side.

Noticed at ISS during installation of the extra-packages.

PR #353, opened 2018-05-24
Move lmfit pin from scikit-beam to pyxrf.

danielballan:
This leaves it up to the individual beamlines whether to pin lmfit to 0.8.3 and
get working xrf code in scikit-beam or accept broken xrf code in scikit-beam in
order to use Larch and modern lmfit.

The ideal fix would be to update the xrf code in scikit-beam, but that update
is not straightforward, and we cannot wait for it.

Closes https://github.com/NSLS-II/lightsource2-recipes/issues/349

Punts on the underlying issue, which is tracked over at scikit-beam:
https://github.com/scikit-beam/scikit-beam/issues/448

### DIFF
--- a/recipes-tag/pyxrf/meta.yaml
+++ b/recipes-tag/pyxrf/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - h5py
     - ipywidgets
     - jupyter
+    - lmfit 0.8.3
     - matplotlib
     - numpy
     - pandas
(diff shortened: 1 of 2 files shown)


danielballan:
P.S. I intentionally did not bump the build numbers here. We need to delete the existing packages and replace them.

This is because of conda behavior that we found surprising when we discovered it. Conda only gives precedence to the highest build number if all else is equal. If it can solve the dependencies more easily by using an old build number, it does so.

danielballan:
I just deleted the old scikit-beam and pyxrf builds to be replaced by this. Hopefully I won the race with the builder. :- )

PR #354, opened 2018-05-24
Bump versions of bluesky, ophyd (prerelease), nslsii

danielballan:


danielballan:
Double-checked: all build numbers are `0`. ;- )

PR #355, opened 2018-05-24
New metapackages for 2018-2.1

danielballan:


recipes-tag/collection/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set year = "2018" %}
 {% set cycle = "2" %}
-{% set version ="0" %}
+{% set version ="1" %}
```
mrakitin: Do you mind to reset the build number to 0?

PR #356, opened 2018-05-24
Reduce the time horizon from 2 days to 30 minutes

mrakitin:
... to avoid overwhelming of freyja

Cron job is currently configured to run every 15 minutes. No updates on freyja needed after merging, the code will be picked directly from GH.


danielballan:
Makes sense to me.

mrakitin:
Thanks, then merging.

PR #362, opened 2018-05-29
fixed csx packages

jrmlhermitte:
fix the csx packages.


mrakitin:
Looks good in general, I hope the build number does not need an update. Let's try that.

jrmlhermitte:
urgh, no it does need an update. ill do it now... thanks for the catch

PR #366, opened 2018-05-29
FIX: remove buggy tests

tacaswell:


mrakitin:
I think it's the last resort. Let's investigate. I've been trying to rebuild the package, but it still fails with segfaults.

mrakitin:
I think we figured it out with @CJ-Wright last time.

PR #368, opened 2018-05-30
Add ipympl.

danielballan:


mrakitin:
I bet it's necessary for Jupyterlab. Merging.

PR #372, opened 2018-05-30
update packages for iss

jrmlhermitte:
fixed up iss packages. currently the latest version needed of isstools is not tagged and needs to be installed after the fact manually. can't tag until i can test and ensure everything works fine


danielballan:
This is a case where you should actually not bump the build number but in fact delete the old packages (first from external, then from internal, so that it doesn't get re-mirrored while you delete). If you just build this as `build: 1` then conda might choose to use `build: 0` instead if doing so reduces the number of things it must install. Build number only take precedence if all else is equal.

jrmlhermitte:
actually this is easier since i should increase the version number. i just bumped it. thanks for the info!

mrakitin:
I think I corrected it. Merging to build and test it at ISS.

PR #375, opened 2018-06-01
Xpd c2 18

CJ-Wright:
These are bug fixes needed for xpd to do calibration.
Attn: @dooryhee


CJ-Wright:
Attn: @jrmlhermitte

jrmlhermitte:
we need pyfai >=0.15.0  for this to work. i can make a new set of yamls later.
there might be no point in testing the build until this is done

CJ-Wright:
I added pyFAI 0.15.0
https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/pyfai/meta.yaml#L2

jrmlhermitte:
it has a few deps that i think are missing in our local repos. we need to add them. i can do this later

PR #376, opened 2018-06-04
fixed path

jrmlhermitte:
conda is looking for the hdf5 plugin in `lib/hdf5/plugin` and not `user/local/hdf5/lib/plugin`

This fixes the issues at CHX. But before we merge, do we know why it was set this way?
@danielballan  maybe this previous path works on a Mac machine? Should we be a little more careful about making a change?


danielballan:
That's a decent guess, but I have no recollection of how this was tested. I think it OK to merge this as long as its works on NSLS-II Debian machines.

jrmlhermitte:
yeah Debian is priority and fortunately, only two beamlines really use this as of now.
id like to hear @mrakitin 's input too, he's been also involved with this at some point.
then we can merge i think. thanks!

mrakitin:
Is it `/lib/...` or conda's `lib/...`? I remember there were commands to set the vars via activate.d and deacrivate.d.

mrakitin:
Please bump the build number in meta.yml if you want it to be built automatically. Otherwise it looks good to me!

tacaswell:
This dos not seem quite right yet.

We also had an issue where if you put the plugin in the same directory as the main hdf5 library you get an infinite recursion.  IIRC, it was dependent on the exact build flags to hdf5 so it may not be a problem any more or it may have been fixed in newer versions of hdf5

This package installs some activate / deactivate scripts which are broken on Jhub because we do not properly activate the enviroment.  I always forget this and may have tested in an activated local enviroment.

If we are putting the libraries in the location where hdf5 searches by default then we can also strip out all of the activate / deactivate logic.

jrmlhermitte:
why are we not properly activating the environment?
by not activating, I'm assuming you mean:
```
/opt/conda_envs/analysis-2018-2.1/bin/python ...
```

as opposed to 
```
conda activate analysis-2018-2.1
...
```


I am :-1: on assuming a conda fixed path. I think either we:
- activate the env before spawning
- put back the fix we removed [here](https://github.com/NSLS-II/playbooks/commit/3331790eb6ddd4bfd22af3dd44f2b629be17e761)

mrakitin:
What's the state of this PR?

ke-zhang-rd:
What's the state of this PR?

mrakitin:
Closing as obsolete. Reopen if needed.

PR #377, opened 2018-06-04
Add FFmepg to standard analysis metapackage.

danielballan:
FFmpeg may be useful at many beamlines, so I propose we include
it in the next roll out (likely 2018-2.2).


danielballan:
Sorry, I missed your comment @mrakitin. I planned for this to go out in this cycle, not a new build of the last cycle. So I think this can be merged.

mrakitin:
Thanks! Looks good then.

PR #381, opened 2018-06-12
HXN: update metapackage

mrakitin:
Found that the deps of HXN's metapackage are pretty outdated. Not sure about hxnfly package, it only exists in [dev channel](https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-dev/hxnfly/meta.yaml), and uses https://code.nsls2.bnl.gov/hg/xf/03id/python/hxnfly as a source.


danielballan:
hxnfly should be kept. Will explain what the deal is on private channels.

mrakitin:
Superseded by #404.

PR #382, opened 2018-06-14
Add xlrd (Excel reading) and xlwt (Excel writing) libs to analysis package

mrakitin:
A few users (including myself) encountered the situation when attempting to read or write Excel files with pandas, but critical dependencies are missing. Adding it here.

Attn @awalter-bnl, @tacaswell.

### DIFF
--- a/recipes-tag/analysis/meta.yaml
+++ b/recipes-tag/analysis/meta.yaml
@@ -38,4 +38,6 @@ requirements:
     - suitcase
     - sympy
     - tornado <5.0.0
+    - xlrd
+    - xlwt
     - xray-vision


danielballan:
There is a separate one for _writing_ that we might as well add while we're here. Not sure what it is called.

mrakitin:
@danielballan, It's called `xlwt`, and also included.

danielballan:
Ah, cool. Somehow I thought it had a longer name. Thanks.

mrakitin:
I didn't bump the build number, so it won't be built this time. Should I have a separate PR for that?

mrakitin:
... and thanks for merging! :)

danielballan:
Bumping the build number doesn't help in cases like this, because conda just ignores it. It only respects the build number if everything else is equal. In this case, it would say, "If I use the old build number I can avoid installing xlrd and xlwt. Win!" Our choices are to delete the old package or wait for the next _version_ bump. We should do the latter.

PR #389, opened 2018-07-25
DEV: no py2

licode:
py36 and py37.


danielballan:
Turning off Python 2.7 build probably should have been a group decision. I think this is _probably_ fine but worth surfacing in a more prominent discussion. attn @NSLS-II/dama

awalter-bnl:
@danielballan There was a generic slack discussion on turning off 2.7 and the group decision was to turn it off. This could have been done more formally however.

danielballan:
Ah, OK. I probably missed it while in vacation mode.

PR #390, opened 2018-07-26
ENH: build with docker-py and Dask

mrakitin:
Uses the following Docker image: https://hub.docker.com/r/mrakitin/debian-with-miniconda/tags/

Requires:
- https://github.com/docker/docker-py
- https://github.com/dask/dask


scripts/nsls2-tag-build.py
```diff
@@ -0,0 +1,53 @@
+import os
+import datetime
+import docker
+from pathlib import Path
+
+
+def run_container(*, pkg_name, pythons=('3.5', '3.6', '3.7')):
```
licode: some document here? Maybe docker_image as input?

scripts/nsls2-tag-build.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Run Docker container
+    docker_client = docker.from_env()
+    output = docker_client.containers.run(docker_image,
+                               name=container_name,
+                               command=f'sh -c "{command}"',
+                               remove=True,
+                               stdout=True,
+                               stderr=True,
+                               volumes=volumes).decode()
```
licode: output to decode?

licode:
good work! some minor suggestion.

scripts/nsls2-tag-build.py
mrakitin: Yeah, I'll document the args. Good idea about the docker image.

scripts/nsls2-tag-build.py
mrakitin: It's needed since the docker function returns a binary string `b''`.

scripts/nsls2-tag-build.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Docker run options
+    docker_image = 'mrakitin/debian-with-miniconda:latest'
+    container_name = f'{pkg_name}-{timestamp}'
+    pythons_str = ' '.join(pythons)
+    command = ['./repo/scripts/build.py',
+               f' /repo/recipes-tag/{pkg_name}',
+               f'-u {upload_channel}',
+               f'--python {pythons_str}',
+               '--numpy 1.14',
```
licode: numpy version as input?

scripts/nsls2-tag-build.py
```diff
@@ -0,0 +1,70 @@
+import os
+import datetime
+import docker
+from pathlib import Path
+
+
+def run_container(*, pkg_name,
+                  docker_image='mrakitin/debian-with-miniconda:latest',
```
mrakitin: Adding this kwarg eventually requires to pass it to the Dask's submitter:
```py
packages = ['bluesky', 'databroker', 'ophyd']
statuses = []
for p in packages:
    st = client.submit(run_container, pkg_name=p, pythons=('3.6',),
                       docker_image='mrakitin/debian-with-miniconda:latest'
                       key=f'{p}-{new_ts()}')
    statuses.append(st)
```

licode: we can ignore it here. or set as default?

mrakitin: No, Dask complains about it if it's not passed, not sure why. Something like:
```py
distributed.worker - WARNING -  Compute Failed
Function:  run_container
args:      ()
kwargs:    {'pkg_name': 'ophyd'}
Exception: TypeError("run_container() missing 1 required keyword-only argument: 'pythons'",)
```

licode: then we can remove docker_image here.

licode:
LGTM. 👍  merge by yourself when you are ready.

mrakitin:
Thanks for the review, @licode!

scripts/nsls2-tag-build.py
mrakitin: Yes, it would be reasonable to have it hard-coded to avoid typing it every time.

PR #395, opened 2018-07-30
FXI: add pypdf2

mrakitin:
Installed manually at the beamline for now (xf18id-ws2), needs to be deployed automatically during the next deployment.


mrakitin:
Thanks for merging, @danielballan!
@gmysage, this will be deployed in the next cycle automatically.

PR #400, opened 2018-08-29
REL: version bump event-model

tacaswell:


mrakitin:
We forgot to change the build number to `0`, I hope it will not cause issues.

PR #406, opened 2018-09-05
event-model: reset build number to 0

mrakitin:
To solve #400.


mrakitin:
Self-merge to test this is built.

PR #411, opened 2018-09-06
REL: Bump collection and analysis packages (2018С3.0)

tacaswell:


recipes-tag/collection/meta.yaml
```diff
@@ -12,9 +12,10 @@ build:
 requirements:
   run:
     - analysis >=2018C2.1
-    - caproto >=0.1.0a1
-    - nslsii >=0.0.4
-    - ophyd >=1.2.0a3
+    - caproto >=0.1.2
+    - event-model >=1.6.0
```
danielballan: No need to re-specify, as this comes in through analysis. Better to specify it in just one place, right?

PR #412, opened 2018-09-06
Move ophyd into 'analysis' metapkg so that bluesky.examples works.

danielballan:


mrakitin:
Do we need to bump a build number here? Or remove the already built packages from anaconda and our local channels.

danielballan:
Let's remove the old ones.

danielballan:
(I'm on it.)

mrakitin:
Sounds good! Merging to trigger new builds.

PR #414, opened 2018-09-07
TST: test ipympl build on freyja and mirroring

mrakitin:
Just a build number bump to test building on freyja since the previous build `0` was created on my laptop. Also, will see if it mirrors this time to pergamon.


mrakitin:
Self-merge to proceed with testing.

PR #415, opened 2018-09-07
FIX: fixing a typo using docker-py

mrakitin:
Not sure how I overlooked it here: https://github.com/NSLS-II/lightsource2-recipes/pull/390/files#diff-f645538a35b21f9b4698a3b9cf8fb816R26.


mrakitin:
Self-merging, tested locally.

PR #416, opened 2018-09-07
WIP: Added fast build bash script using mrakitin/debian-with-miniconda

mrakitin:
I realized we never switched to this faster version of the building approach, which was superseded by #390. To have a history of changes, I'd like to keep it here, even if it won't be used.


mrakitin:
We have a working Python script for this. Closing in favor of it.

PR #420, opened 2018-09-11
Add inflection package for BMM

mrakitin:
Copied from https://github.com/conda-forge/inflection-feedstock/blob/master/recipe/meta.yaml.


danielballan:
@mrakitin Would you add the new BMM metapackage to the `extra_packages` configuration parameter in the NSLS-II/playbooks `production` inventory? Let me know if I should explain more! :- D

mrakitin:
Thanks for reviewing and merging, @danielballan! Do we care it's a `noarch` package?

danielballan:
That's fine, assuming that the conda-forge package is correct.

PR #426, opened 2018-09-13
2018C3.0 for csx1

licode:


danielballan:
Are those new _direct_ dependencies at CSX? If they are actually secondary dependencies, pulled in through hklpy, it would be better to update the hlkpy recipe and then bump the hlkpy pin here.

danielballan:
To explain why a bit more: by updating hlkpy itself, these updated requirements will automatically propagate to _all_ beamlines that need it, rather than requiring us to keep many metapackages in sync manually.

licode:
those are deps on hkl. Basically no change needed here I think.

danielballan:
It would be best if we could pin this such that we are sure that we get a build of hklpy that has the seconardy deps that we want. But we can circle back to that question once you have update the hklpy recipe.

licode:
close it for now. reopen when hkl is ready.

PR #427, opened 2018-09-13
MNT: bump tomopy build number

tacaswell:
Something in this is pinning to an old version of readline which is
preventing it from being installed with the 2018-3.1 environment.


tacaswell:
Thanks @danielballan

PR #430, opened 2018-09-13
MNT: python is a runtime dep of nslsii

tacaswell:


danielballan:
Same question. Surely we get Python through `numpy` or something else. How did this come up?

tacaswell:
To update publicly what the issue is:  If python is not a runtime dep, then `py3X` is not added to the tarball file name, which means the 3.6 and 3.7 clobber each other.  We were left with a race condition (not sure if first one wins or last one wins) where depending on the order we build these packages in we would get identically named files.  If the 3.7 build 'won' then it would happily install into a 3.6 enviroment but not be importable (because it was in PREFIX/lib/python3.7/site-packages).

PR #433, opened 2018-09-18
Add slacker package

mrakitin:
**Source:** https://github.com/conda-forge/slacker-feedstock/blob/master/recipe/meta.yaml
**Motivation:** remove conda-forge dependency from https://github.com/NSLS-II/debian-with-miniconda/pull/6/files#diff-3254677a7917c6c01f55212f86c57fbfR75


mrakitin:
Thanks Tom!

PR #435, opened 2018-09-19
Fixes to build tomopy

mrakitin:
This fix solves the following:
- the wrong/updated sha256 of the tomopy recipe
- the h5py package, which should come from the `anaconda` channel, not the `lightsource2-tag` one


mrakitin:
Thanks @licode!

PR #443, opened 2018-09-27
No noarch

CJ-Wright:
Maybe this will fix the segfault?


tacaswell:
any theory as to why noarch would be related to segfaults?

CJ-Wright:
I'm more worried by the python constraint than the noarch, but no. I haven't been able to reproduce the error on CF (or my local) so this is a bit of a shot in the dark (although I think this recipe is one of the few to actually restrict the python version with noarch so :man_shrugging: ).

recipes-tag/xpdacq/meta.yaml
```diff
@@ -39,8 +39,8 @@ requirements:
 test:
   imports:
     - xpdacq
-    - xpdacq.xpdacq
-    - xpdacq.beamtime
+#    - xpdacq.xpdacq
+#    - xpdacq.beamtime
```
mrakitin: ha, that will definitely help since it's failing on the tests.

CJ-Wright: Exactly :)

mrakitin:
OK, let's retry it.

mrakitin:
For completeness - it was built successfully: https://anaconda.org/lightsource2-tag/xpdacq/0.8.2/download/linux-64/xpdacq-0.8.2-py36_0.tar.bz2.

PR #444, opened 2018-09-27
Add conda-execute

mrakitin:
We have a manually-mirrored package in https://anaconda.org/lightsource2-tag/conda-execute/files.  Need to have a recipe for that.

Cloned from https://github.com/conda-forge/conda-execute-feedstock/blob/master/recipe/meta.yaml.


mrakitin:
Self-merging. Let's see how it builds.

PR #445, opened 2018-09-27
add back tests

CJ-Wright:


mrakitin:
Based on success in #443, let's be optimistic ;)

PR #365, opened 2018-05-29
update xpd and pdf

jrmlhermitte:

PR #446, opened 2018-09-27
PDF: fix metapackages

mrakitin:
#442 was never built since the package name is wrong (overlooked in #365).

It's being built now.


mrakitin:
@CJ-Wright, that should have been merged before #474. Please merge so that we could proceed with correct names of the PDF packages.

mrakitin:
Thank you!

PR #447, opened 2018-09-27
CLN: remove h5py recipe to avoid internally built 2.8.0

mrakitin:
Attn @tacaswell


mrakitin:
I removed the erroneously built h5py 2.8.0 from https://anaconda.org/lightsource2-tag/h5py/files?sort=basename&sort_order=desc and alexandria.

licode:
I think we are using 2.9 now from -tag. And there is 2.8 from internal defaults. So it should be safe to remove this 2.8.

mrakitin:
I think I just close this PR as it removes the files which we still need.

PR #451, opened 2018-10-12
Add zsh package

mrakitin:
Needed for swift-t.

PR #454, opened 2018-10-12
zsh: fix notty issue

mrakitin:
This fixes the issue https://github.com/NSLS-II/lightsource2-recipes/pull/451#issuecomment-429365550.

Related [documentation](https://github.com/zsh-users/zsh/blob/3d7215cc8277b39cd1e24ce5a04376d45bfbabf0/INSTALL#L569-L574):
```
A test for the function tcsetpgrp is turned on by default.  The test
needs to run the function to determine if the implementation is
usable. However, this can cause problems when configure is run without
a controlling terminal (eg. from cron).  To avoid this, use
--with-tcsetpgrp or --without-tcsetpgrp to tell configure whether the
function should be used.
```


mrakitin:
This is already built locally and pushed to https://anaconda.org/lightsource2-tag/zsh/5.6.2/download/linux-64/zsh-5.6.2-0.tar.bz2.

licode:
OK, but we can still merge it here.

mrakitin:
Absolutely! Thanks!

tacaswell:
why are we packaging zsh?!

tacaswell:
I see, weird that it depends on a shell...

mrakitin:
Yep. http://swift-lang.github.io/swift-t/guide.html#_prerequisites (item 3) explains the requirements:

> Ensure you have these third-party software development tools (APT packages are denoted in parentheses): SWIG (swig), ZSH (zsh), Apache Ant (ant), a Java Development Kit (default-jdk), Make (make), and GCC for C (gcc).

mrakitin:
This script checks the deps are installed: https://github.com/swift-lang/swift-t/blob/e9579d01588e72656c7e8e7f07575b0f3da056ab/dev/build/check-tools.sh#L8.

And the main turbine script is written in zsh: https://github.com/swift-lang/swift-t/blob/e9579d01588e72656c7e8e7f07575b0f3da056ab/turbine/code/bin/turbine.in#L1

PR #455, opened 2018-10-12
ISS: add xraylarch to the metapackage

mrakitin:


mrakitin:
Self-merging it since I'd like to push new environments to ISS now (after it's built).

mrakitin:
I didn't consider #372. Will have to create a separate PR and close the old one.

PR #456, opened 2018-10-12
hdf5-lz4: correct conda variable of the active env

mrakitin:
Observed this issue at the HXN lab during today's visit. Fixed by it.

More details are at: https://stackoverflow.com/a/42660674/4143531.


mrakitin:
Haha, noticed our previous conversations about the changed var: https://github.com/NSLS-II/lightsource2-recipes/pull/226. So, the evolution is like that:
`CONDA_ENV_PATH` -> `CONDA_PREFIX` -> `CONDA_DEFAULT_ENV`

mrakitin:
#376 has some relevant discussions, but is not about the var itself. Just wanted to mention it in this PR.

licode:
so conda_prefix is the full path?

mrakitin:
No, it wasn't defined at all on that machine. It looks like conda env vars changed again recently, or, at least, are different on Mac and Linux. `CONDA_DEFAULT_ENV` should be used now.

danielballan:
Thanks for cross-referencing the old issues and summarizing the evolution here!

PR #459, opened 2018-10-17
Add rixs and sixtools

mrakitin:
`rixs` should be built first as `sixtools` depends on that.


mrakitin:
Thanks, @awalter-bnl!

PR #461, opened 2018-10-19
nslsii: bump version v0.0.7

mrakitin:
https://github.com/NSLS-II/nslsii/releases/tag/v0.0.7


mrakitin:
Observed issues with building conda packages, which appeared on previously working Docker images. It turned out to be a major upstream issue: https://github.com/conda/conda/issues/7870.

Recommendation in https://github.com/conda/conda/issues/7870#issuecomment-431519432 was to avoid using `anaconda` channel. The `defaults` channel should be used instead.

licode:
thanks for the quick followup.

PR #462, opened 2018-10-26
Add packages for ISS

mrakitin:
This also adds more tweaks to `build.py` and `nsls2_tag_build.py`.

### Sources:
- [pyfiglet](https://github.com/conda-forge/pyfiglet-feedstock/blob/master/recipe/meta.yaml)
- [mendeleev](https://bitbucket.org/lukaszmentel/mendeleev/src/fcbaa655eb406237ae465ccc44436248d4b8d145/conda-recipe/meta.yaml?at=default&fileviewer=file-view-default)
- [periodictable](https://github.com/conda-forge/periodictable-feedstock/blob/master/recipe/meta.yaml)


mrakitin:
Ready for review. Packages have been built and pushed to our channels (but not to the beamline).
Attn @elistavitski

recipes-tag/08-id-iss-analysis/meta.yaml
```diff
@@ -1,17 +1,19 @@
...
-{% set version ="0" %}
+{% set version ="1" %}
 
 package:
   name: 08-id-iss-analysis
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 0
+  number: 1
```
danielballan: Should this be `0`?

recipes-tag/08-id-iss-collection/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
-{% set version ="0" %}
+{% set version ="1" %}
 
 package:
   name: 08-id-iss-collection
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 2
+  number: 1
```
danielballan: and this also?

recipes-tag/08-id-iss-analysis/meta.yaml
mrakitin: If I remember correctly, I did a few builds aiming to resolve some dependency issues, that's why there are numbers > 0. See https://anaconda.org/lightsource2-tag/08-id-iss-analysis/2018C3.1/download/linux-64/08-id-iss-analysis-2018C3.1-1.tar.bz2.

recipes-tag/08-id-iss-collection/meta.yaml
mrakitin: See above, the file is already built and is at our public channel: https://anaconda.org/lightsource2-tag/08-id-iss-collection/2018C3.1/download/linux-64/08-id-iss-collection-2018C3.1-1.tar.bz2.

mrakitin:
Thanks for reviewing, @danielballan! I should have deleted broken/invalid builds with the build number `0`, and rebuilt everything. Next time it will be cleaner.

recipes-tag/08-id-iss-analysis/meta.yaml
danielballan: But you're bumping the version number, so you can reset the build number to 0, right?

mrakitin: The situation was as follows:
- I built the packages with the build number `0`, and they were pushed to anaconda and mirrored to nsls2-tag.
- I discovered some other packages to add, and then, withing the same PR, bumped the build number, and built with the build number `1`, to have them available for the push to the beamline.

You convinced me that it's better to have a clearer numbering, so I removed the packages from anaconda and our local mirror, and reset the build numbers to `0` in 87b5444, so that we have proper packages.

danielballan:
Now I understand. Thanks for cleaning it up. As always, I worry that conda's feature of not giving priority to build numbers might bite us.

mrakitin:
@danielballan, I don't fully trust it either (see multiple pyfai build attempts). Thanks for merging!

PR #466, opened 2018-11-06
Fix ipy* package versions for JupyterHub to work correctly

mrakitin:
Resolves https://github.com/NSLS-II/playbooks/issues/159.

Also, incremented the version of the collection/analysis packages to `2018-3.3`. Please note, we didn't release `2018-3.2`, but ideally it would be good to have a matching version of the package with the new envs (see https://github.com/NSLS-II/playbooks/pull/161).


recipes-tag/analysis/meta.yaml
```diff
@@ -25,9 +25,10 @@ requirements:
     - ffmpeg
     - graphviz
     - h5py >=2.8.0
-    - ipympl
+    - ipykernel==4.8.2
```
danielballan: This one doesn't affect widgets, so I think we should leave it unpinned.

mrakitin: OK, thanks. Unpinned in https://github.com/NSLS-II/lightsource2-recipes/pull/466/commits/80368470b24ad17822a0072a6a8786c560e6f755.

mrakitin:
Thanks for reviewing and merging, @danielballan!

PR #467, opened 2018-11-06
Another attemp to build pyfai

mrakitin:
Let's try that, @CJ-Wright.

(also, sorted the dependencies)


mrakitin:
Thank you, @CJ-Wright!

PR #468, opened 2018-11-06
ipympl v0.1.1 to work with JH

mrakitin:
It's built already to be available for #466.


mrakitin:
Self-merge as a part of work for #466.

PR #471, opened 2018-11-16
Revive HKL

mrakitin:
This was built successfully using https://hub.docker.com/r/mrakitin/debian-with-miniconda/tags/ (20181116161941) and the existing recipe for `hkl`.


licode:
If you check dependencies, I think this building of hkl uses pygobject3 ==3.18.2 and glib=2.51. But all our stack is based on glib=2.56, that's why this hkl can't be installed.

mrakitin:
Let's try a newer version of pygobject v3.31.1, which has a different installation method (`python setup.py install`). I built is successfully locally, it's pushed to https://anaconda.org/lightsource2-tag/pygobject/3.31.1/download/linux-64/pygobject-3.31.1-py36_0.tar.bz2.

licode:
great. let's have a try! new version may help.

licode:
good job. The new version is only there for 2 days. http://ftp.acc.umu.se/pub/GNOME/sources/pygobject/

Let's see.

PR #482, opened 2018-12-18
Add pyfftw to the analysis package for deployment during the 19-1 cycle

### LINKED ISSUE #480: Add pyfftw to facility-wide analysis package
Seems likely to be widely useful, has been incidentally requested at various times by a couple beamlines now.

mrakitin:
Closes #480.


danielballan:
If I remember right, we want to bump the version only once all of the updates for 2019C1 have been made so that the package will be built once. (Last time we bumped too early and had to go through the hassle of deleting the built packages from the two servers.)

mrakitin:
I had this thought as well and hesitated to bump the version. Now I checked out to 8930e91158a05f7d9396bf6cad72283878446992 and force-pushed, so there is just the package added without bumping the version.

PR #484, opened 2019-01-03
h5py 2.9.0 from conda-forge

licode:


licode:
sounds good. we may also want to remove the old version of 2.6.0? defaults also has 2.6.0, so no need to keep extra copy.

mrakitin:
Great! Yes, please remove 2.6.0, so we are not confused by this artificial versioning on top of git 😃

recipes-tag/h5py/run_test.py
```diff
@@ -19,7 +19,6 @@
 import h5py.h5t
 import h5py.h5z
 import h5py.utils
-assert h5py.version.hdf5_version_tuple == (1, 10, 1)
```
mrakitin: Do you know why it's deleted? Was it our channel's modification to check for the strict hdf5 version?

licode: I run some tests, so hdf5 version can be 1.10.0 or 1.10.2 or higher. I didn't see restriction on this. 

I just copied from conda-forge, and didn't make any change.

licode:
one reason for those old pkg folders is because defaults channel is kind of slow for new versions. So we try conda-forge recipes locally to build it. but later, we need to come back to remove it as it is ready in defaults.

recipes-tag/h5py/run_test.py
mrakitin: OK, sounds good.

PR #486, opened 2019-01-03
packaged ptychography software (frontend + backend)

leofang:
I finally packaged the ptychography code deployed at HXN and CSX. There're things to be discussed offline with @licode and perhaps @stuartcampbell before merging. Among those, the usage of `git submodule`, currently hosted in my private repo in order to avoid patent issues, may be an obstacle for the building system? Perhaps moving both repos (GUI and GPU code) to NSLS-II and allowing read-only access for the building agent is the best option? (We need to grant people from other nat'l labs read-only access anyway.)


mrakitin:
Thanks @leofang.

Did you try to build it locally with `conda build .` in the recipe folder? You can use a Docker image from https://cloud.docker.com/u/nsls2/repository/docker/nsls2/debian-with-miniconda or https://cloud.docker.com/u/mrakitin/repository/docker/mrakitin/debian-with-miniconda to try it - they have preinstalled conda and necessary packages.

Also, is https://github.com/leofang/ptycho_gui the private repo you are talking about? I think I can access it. You probably meant your *personal* repo. I checked the structure of the project briefly, and it seems it won't install as the setup.py is missing. You can consider https://github.com/NSLS-II/scientific-python-cookiecutter to prepare a proper structure of the project to make it pip-installable.

recipes-tag/nsls2ptycho/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - mpi4py 
+    - qt >=5
+    - pyqt >=5
+    - numpy
+    - scipy
+    - matplotlib
+    - pillow
+    - h5py
+    - posix_ipc
+
```
licode: it is better to have some import test here.

leofang:
> Did you try to build it locally with `conda build .` in the recipe folder? 

Yes, I did `conda build --python 36 nsls2ptycho` and then `conda install --use-local nsls2ptycho`. After that I tested the installed package works expected.

> You can use a Docker image from https://cloud.docker.com/u/nsls2/repository/docker/nsls2/debian-with-miniconda or https://cloud.docker.com/u/mrakitin/repository/docker/mrakitin/debian-with-miniconda to try it - they have preinstalled conda and necessary packages.

Sorry I don't understand the reason for using a Docker image, @mrakitin could you please explain?

> Also, is https://github.com/leofang/ptycho_gui the private repo you are talking about? I think I can access it. You probably meant your _personal_ repo. I checked the structure of the project briefly, and it seems it won't install as the setup.py is missing. You can consider https://github.com/NSLS-II/scientific-python-cookiecutter to prepare a proper structure of the project to make it pip-installable.

The project has two parts: a frontend (a GUI) which is https://github.com/leofang/ptycho_gui as you noticed, and a backend (CPU+GPU codes) which is hosted in private mode https://github.com/leofang/ptycho. The backend is set to be a `git submodule` of the frontend, see the `ptycho` folder in https://github.com/leofang/ptycho_gui/tree/packaging/nsls2ptycho/core.

Ah, I see the confusion. The `packaging` branch is pip-installable, which is what I specified in the recipe. I'll merge it to `master` later. When you clone the frontend repo, in the source you can do `pip install .` I made sure of this before moving to write a conda recipe based on `pip install .` Does this make sense to you?

Thanks!

leofang:
Great, thanks @mrakitin @tacaswell. I built the package on `xf03id-srv5` since they have GPUs that I can test right after building and installing, but I'll try using docker and report here later.

recipes-tag/nsls2ptycho/meta.yaml
leofang: test for the dependencies or the package itself?

leofang: two import tests added

leofang:
I've followed @mrakitin's instruction to build with docker and it worked after a few slight changes in `meta.yaml` and `setup.py` (mainly to address "numpy header not found"). I also added two import tests as suggested by @licode. 

The only thing left before making all these work is to mirror the backend to the internal GitLab server so that the builder can access it, but I don't know how to do this.

leofang:
@mrakitin the backend has been moved to internal GitLab, and a local build was successful. However, during building I still need to type my GitLab id and pw because the repo's visibility is set to "internal". Any idea to get around this without making the repo public?

mrakitin:
@leofang, do you have a [GitLab token](https://docs.gitlab.com/ee/user/project/deploy_tokens/)? Maybe we can use it to authenticate and clone the code from your private repo?

leofang:
I saw that page too. But I think our internal GitLab is not on the latest version, so it lacks many features including deploy tokens. This is [what I see in Settings->Repository](https://www.dropbox.com/s/uxopd47ulu2vll8/%E8%9E%A2%E5%B9%95%E5%BF%AB%E7%85%A7%202019-03-23%20%E4%B8%8B%E5%8D%881.52.41.png?dl=0).

leofang:
I thought about it, and think if the internal GitLab has no particular advantage for us, I can just create a token in my private GitHub repo and use it. Perhaps this will suffice?

leofang:
I took some time to investigate, and my understanding is 
1. Both GitHub and GitLab support repo-level deploy keys (ssh keys) and user-level tokens (called Personal Access Token). 
2. With a user-level token, creating a git submodule using a remote url like `https://leofang:TOKEN@github.com/leofang/ptycho.git` will make `git clone` work with my private repo.  
3. However, the token gives full access to *everything* under the user's account, and since at the submodule level the `TOKEN` is stored as plain text, we don't really wanna expose it as it's unsafe.
4. So we wanna use deploy keys. I tested that if we create a ssh key pair in the docker image, copy the public key to my private repo as a deploy key, and in the git submodule use a url `https://leofang@github.com/leofang/ptycho.git` (no need to use `TOKEN` this time), `git clone` will also work. 

So, my new question is: is it unsafe to store the ssh key pair from `ssh-keygen` in a docker image? If it's ok, then we can just update the docker image and everything else will work.

mrakitin:
Great job on investigating the options! It is save to store .pub part of the ssh key pair, but never the private pair. What we do for other tokens is passing them via env variables to the Docker container when we start it. That way we don't have to worry about storing them in the Docker image. If you want to use the ssh pair, we can mount it if the secret dir exists on our buildind server (e.g., `/home/youruser/secrets/` --> `/secrets/`).

leofang:
> If you want to use the ssh pair, we can mount it if the secret dir exists on our buildind server (e.g., /home/youruser/secrets/ --> /secrets/).

Could you clarify this? What you want to mount? What is to be put in the `secrets` dir?

mrakitin:
If you place your public/secret keys somewhere on `freyja`, we can mount docker image with the `-v`  option (see docs [here](https://docs.docker.com/storage/bind-mounts/)). We already mount the cloned repo here: https://github.com/NSLS-II/lightsource2-recipes/blob/b1e0544f1b7c2e33cc24bd2aeed01855883c88aa/scripts/nsls2-tag-build.sh#L37
We can do something similar with your keys.

As this package is very special, for now I think we will be fine with accepting a built package from you, which can be put to our local channel on `alexandia` manually without external mirroring. You could provide a ready package (`tar.bz2`), I can place it to the local channel.

leofang:
@mrakitin I gave up ssh for a number of reasons that we can discuss offline. Now I switch back to personal access tokens in f8878ad. The key insight is that `https://TOKEN@github.com/leofang/ptycho_gui.git` is a valid git url (no user name is needed! FYI, this doesn't work for GitLab.) and `TOKEN` will propagate all the way to the submodules, therefore it's not needed to expose it as plain text anywhere. 

I tested in docker and it works, so we're only one step away: Is it possible to set an environmental variable named `GH_TOKEN` before the build starts? (I store my token on Freyja so it should be accessible there.) 

> As this package is very special, for now I think we will be fine with accepting a built package from you, which can be put to our local channel on alexandia manually without external mirroring. You could provide a ready package (tar.bz2), I can place it to the local channel.

OK. This will be the last resort.

Let me know what you think. Thanks!

mrakitin:
Great progress, @leofang! The problem now is that freyja will push every built package to our public channel; we don't have the machinery not to push (yet). I think it will need some preparation before we can proceed. On the package side, we may need some flag to mark the package as "private".

leofang:
Ah, that's right. We discussed this long ago on Slack. Glad we're finally here! I suppose this could be a good summer project? 😆 

So for the time being let's use the last resort. I think when you manually place the package to the local channel, it won't be uploaded, right?

mrakitin:
> Ah, that's right. We discussed this long ago on Slack. Glad we're finally here! I suppose this could be a good summer project? 😆

Our building approach will be significantly reworked. I guess freyja will only be needed for such corner cases.
 
> So for the time being let's use the last resort. I think when you manually place the package to the local channel, it won't be uploaded, right?

Correct, the mirroring only happens in one direction: https://anaconda.org/lightsource2-tag/ --> alexandria.

mrakitin:
Here is a summary about gitlab ssh access: https://3.basecamp.com/3733838/buckets/4534819/messages/1693665280.

leofang:
Just leaving a note for myself when I return to this PR in the near future.

Resolving the CUDA dependency issue for Conda packages is still challenging. One of the reasons is that currently `cudatoolkit` is an incomplete package that only contains runtime libraries and lacks the compiler `nvcc` and friends. While it is impossible at the Conda level to install the CUDA driver (which requires root), it should be possible for Conda to infer the driver version and install the toolkit with the highest version supported by the driver. Indeed, this is an ongoing effort that will be included in the latest Conda release, see discussions in conda-forge/conda-forge.github.io/issues/687 and conda/conda/pull/8267. Hopefully, we will be able to add something like `- cudatoolkit >=9.0` in the `build` section of the recipe. (And as users, `conda install cudatoolkit=9.0`) Then, we can work toward packaging ptychography, tomography, and more software that offer GPU support more easily.

For this particular package, in [this branch](https://github.com/leofang/ptycho_gui/tree/cuda_build) I've resolved the GPU & CuPy issues at the `pip` level by assuming GPU may not exist. If detected, `nvcc` will be invoked to generate CUDA binaries that will be `pip install`ed, and the corresponding version of CuPy will be added as a dependency (if not already installed). So, once the new `cudatoolkit` recipe is ready, at the Conda level we will be able to provide a `nvcc` (regardless of whether it's already available or not) by adding `cudatoolkit` to the `build` section as mentioned above so that the `pip` solution will always have a `nvcc` to work with.

leofang:
`recipes-private` sounds better as it indicates the recipes in this folder won't (and can't) go public. If it's OK to you I'll proceed.

mrakitin:
Yes please, let's do it. This way we may have subsequent PRs to that folder rather than keeping iterating in this single WIP PR.

leofang:
Done. Question: if this PR is merged now, would it trigger the builder (I hope not)?

mrakitin:
Thanks Leo. It won't trigger a build as it only tracks the changes in `recipes-tag`: https://github.com/NSLS-II/lightsource2-recipes/blob/c8c79b368416388db1f1d0a22aaa05f5408e3268/scripts/nsls2-tag-build.sh#L22
and the building itself is happening here: https://github.com/NSLS-II/lightsource2-recipes/blob/c8c79b368416388db1f1d0a22aaa05f5408e3268/scripts/nsls2-tag-build.sh#L75

recipes-private/nsls2ptycho/build.sh
```diff
@@ -0,0 +1,19 @@
...
+# need the CPU version, then we don't need to touch anything GPU-related!
+#
+# For MPI (and mpi4py), the problem is to build mpi4py on top of whatever
+# MPI in the system. Conda is unable to do a customized build, and it's 
+# better to install a specified MPI first (either from Conda or by oneself,
+# if, say, cuda-aware MPI is needed), and then pip install mpi4py.
+#
+# Currently this script assumes 1. CuPy is either not needed or installed 
+# separately, and 2. "conda install mpi4py" works. Obviously we need to 
+# polish this when more beamlines need this pacakge...
```
mrakitin:
```suggestion
# polish this when more beamlines need this package...
```

recipes-private/nsls2ptycho/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+$PYTHON -m pip install . --no-deps --ignore-installed -vv
```
mrakitin: If it's the only command needed, it can be handled via `meta.yml`. The comments are valuable, but can be in a `README.txt` or something similar...

recipes-private/nsls2ptycho/meta.yaml
```diff
@@ -0,0 +1,52 @@
+{% set token = environ.get('GH_TOKEN') %}
```
mrakitin: I like this trick! If I use my token, I guess it won't work as I don't have access to your repo, right? How can we make it more sustainable?

recipes-private/nsls2ptycho/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+package:
+  name: nsls2ptycho
+  version: "1.1.1"
+
+source:
+  git_rev: v1.1.1
+  git_url: https://{{ token }}@github.com/leofang/ptycho_gui.git
```
mrakitin: This repo seems to be public. Why do we need the token? How is the private repo with the backend (https://github.com/leofang/ptycho) used?

recipes-private/nsls2ptycho/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - pyqt >=5
+    - numpy
+    - scipy
+    - matplotlib
+    - pillow
+    - h5py
+    - posix_ipc
+    - databroker
+    - attrs >=17.4.0 # needed because anaconda messed up the pytest dependency...see pytest-dev/pytest/pull/3228
+    - numba >=0.41.0
```
mrakitin: We tend to sort the dependencies alphabetically. Why not to do it here? :)

recipes-private/nsls2ptycho/meta.yaml
leofang: If I add you to the repo, you'll have access. The only drawback I encounter for now is that if we forget to set the env variable, the error message is obscure :D  

For the long term, I am hoping that @stuartcampbell will allow both the frontend and the backend repos to be transferred to NSLS-II so that we can assign team privileges. For example, DAMA will have the write access, and collaborators will have read-only (or write if they want to contribute) access.

recipes-private/nsls2ptycho/meta.yaml
leofang: Because the private repo is made as a submodule of the public one, see here:
https://github.com/leofang/ptycho_gui/tree/master/nsls2ptycho/core
When building the package, `git clone --recursive` is done so that both repos are accessed, and the token is then used.

recipes-private/nsls2ptycho/build.sh
leofang: Fixed, thanks.

recipes-private/nsls2ptycho/meta.yaml
leofang: Done.

recipes-private/nsls2ptycho/build.sh
leofang: This belongs to a discussion with a bigger scope, maybe worthy of a separate issue. I tend to leave the script here until I (or we) find a satisfying solution for the dependency issue. 

First, non-Python libraries (ex: HDF5, FFTW) can be built against MPI to provide asynchronous processing. Then, the corresponding Python wrappers (ex: h5py, PyFFTW) need to be built against those specialized versions. It means at the Conda level the complexity inflates quickly, and I am not sure what is the best strategy (use build string as unique id? use different output names?) to handle this.

Second, when it comes to supporting (NVIDIA) GPUs, even MPI itself can have different flavors, so called CUDA-awareness. Then, in addition to building mpi4py against a CUDA-aware version, for example, do we build HDF5 against a CUDA-aware MPI or not? Too many degrees of freedom come into play, and so far we only keep the minimum requirement. 

I feel that eventually a Bash or Python script will be needed to help Conda resolve this issue, especially in the coming Jupyter-SDCC era. The `meta.yaml` recipe alone might not be enough.

mrakitin: OK, sounds good. Let's keep it as a placeholder for the future complications then.

recipes-private/nsls2ptycho/meta.yaml
mrakitin: Since it's a manual process of building anyway, it's OK. Having a broken environment in a container is safe, and the package won't be uploaded, so I'm not concerned much.

Could you please add me as a collaborator on that private repo before we move the repo(s) to the new home? Thanks.

recipes-private/nsls2ptycho/meta.yaml
mrakitin: Oh, right! I remembered it was a submodule, but didn't see it in the root of the repo.

recipes-private/nsls2ptycho/meta.yaml
leofang: Done.

leofang:
:beer:

mrakitin:
🍻

PR #493, opened 2019-01-04
MNT: try using the defaults channel rather than anaconda channel

tacaswell:


licode:
can you also change it at nsls2-dev-build.sh  and nsls2-tag-build-local.sh to be consistent?

tacaswell:
done and force-pushed

licode:
thanks!

PR #494, opened 2019-01-04
DEV: no limit on conda version

licode:


danielballan:
Seems like a good idea in principle but do we want to do this _now_ or wait until the current flurry of activity is done?

I believe we pinned this to avoid the issues around code moving between conda and conda-env, generating some incompatible version pairs.  By now it _should_ be safe to unpin. Unpinning conda build may have the effect of changing the build string, which could affect mirroring via its deduplication step.

mrakitin:
It's rather a conflicting change with what we do in https://github.com/mrakitin/debian-with-miniconda/blob/32a417ae68e69c39a0b4bebbb15a5abfacc83cc3/Dockerfile#L101. I'd prefer to have the update in Docker rather than here. Postponing the decision till July's deployment.

ke-zhang-rd:
Could I merge this?

mrakitin:
This is a conflicting PR, I think we have what we need in master already. Closing.

PR #496, opened 2019-01-04
DEV: csx meta pkgs

licode:


recipes-tag/23-id-1-csx1-analysis/meta.yaml
```diff
@@ -11,4 +11,4 @@ build:
```
mrakitin: reset build # to 0?

recipes-tag/04-id-isr-collection/meta.yaml
```diff
@@ -11,9 +11,7 @@ build:
```
mrakitin: Consider changing the version to `2019-1.0` for this package as well.

recipes-tag/04-id-isr-collection/meta.yaml
```diff
@@ -11,9 +11,7 @@ build:
 
 requirements:
   run:
-    - gsl
-    - libiconv
-    - pcre
     - hkl
```
mrakitin: I don't think we need hkl explicitly here. It's taking care of by hklpy itself: https://github.com/NSLS-II/lightsource2-recipes/blob/5e3f1850af2a45ba011e1b554d4318510871beb3/recipes-tag/hklpy/meta.yaml#L21

recipes-tag/04-id-isr-collection/meta.yaml
```diff
@@ -11,9 +11,7 @@ build:
...
-    - hklpy
+    - hklpy >=0.3.13
     - eiger-io
```
mrakitin: `eiger-io` will come from `04-id-isr-analysis`, no need to keep it.

recipes-tag/04-id-isr-collection/meta.yaml
```diff
@@ -11,9 +11,7 @@ build:
...
+    - 04-id-isr-analysis
```
mrakitin: consider sorting the dependencies alphabetically.

recipes-tag/23-id-1-csx1-collection/meta.yaml
```diff
@@ -11,10 +11,7 @@ build:
...
 requirements:
   run:
-    - gsl
-    - libiconv
-    - pcre
     - hkl
-    - hklpy >=0.3.12
+    - hklpy >=0.3.13
     - 23-id-1-csx1-analysis
     - hxntools
```
mrakitin: sorting would not hurt here as well.

recipes-tag/23-id-1-csx1-collection/meta.yaml
```diff
@@ -11,10 +11,7 @@ build:
 
 requirements:
   run:
-    - gsl
-    - libiconv
-    - pcre
     - hkl
```
mrakitin: no need of `hkl` as it's a dependency of `hklpy` (see above).

recipes-tag/23-id-1-csx1-collection/meta.yaml
```diff
@@ -11,10 +11,7 @@ build:
```
mrakitin: build # reset to 0.

licode:
good catch, thanks!

recipes-tag/23-id-1-csx1-analysis/meta.yaml
```diff
@@ -1,14 +1,14 @@
...
   version: {{ year }}C{{ cycle }}.{{ version }}
 
 build:
-  number: 1
+  number: 0
 
 requirements:
   run:
-    - csxtools
+    - csxtools 0.1.13
```
danielballan: Do we need to pin it to _exactly_ this version or only at least this version?

danielballan:
My review is too late for the merge, but I think it might flag an important issue worth a follow-up PR, @licode.

recipes-tag/23-id-1-csx1-analysis/meta.yaml
licode: I prefer to use stable version 0.1.13 at this moment. I remember I saw bugs in 0.1.14, but didn't get time to follow up. I may try to update to 0.1.14 during mid-cycle after I am sure 0.1.14 is safe to go.

PR #498, opened 2019-01-07
pin netcdf4 to hdf5=1.10.0

licode:
local test works.


licode:
copy from conda-forge

recipes-tag/netcdf4/.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,15 @@
+Thanks for your interest in helping out conda-forge.
```
mrakitin: @licode, do we actually need this directory? Same for `.circleci` above.

licode: thanks for correcting. But this repo also needs to be removed as it builds against wrong hdf5 version of 1.10.0. We have new version against on hdf5 1.10.2 in defaults

PR #502, opened 2019-01-08
MNT: add dependency on python

tacaswell:
This is critical so the pyXY goes in the name.

This should eventually be a no-arch package, but there are other
standing issues with that for now.


danielballan:
This won't work. The conda solver will say, "I can use build 0 or build 1, and I see that build 0 has fewer deps so let's use that." The old build has to be removed before this is merged.

tacaswell:
can we merge this so it gets built and deal with removing it in paraller?

PR #505, opened 2019-01-08
Update CHX metapackages

mrakitin:
That should be built after the packages from #503 are built.


mrakitin:
The packages from https://github.com/NSLS-II/lightsource2-recipes/pull/503 were built and mirrored. Self-merging to make it going.

PR #506, opened 2019-01-08
update pyfai to 16

CJ-Wright:
This builds on CF on gcc7 (not toolchain, gcc4)


tacaswell:
Lets see if this builds internally....

PR #507, opened 2019-01-08
pin hdf5=1.10.2

licode:


recipes-tag/h5py/meta.yaml
```diff
@@ -9,7 +9,7 @@ source:
   sha256: 49d593c07a3a4cb87ed421f502e34368b98282a4ac6d4388379aab80b11c7924
 
 build:
-  number: 1000
+  number: 0
```
tacaswell: Is this going to collide with an existing build number?  We started on 1000, maybe this should be 1001?

licode:
self merged.

PR #512, opened 2019-01-14
MNT: Remove pychx

ke-zhang-rd:


danielballan:
For context, the rationale here is that CHX generally needs a source-installed version, so we shouldn't attempt to manage the package here (yet).

PR #527, opened 2019-01-28
Mpl modest image

mrakitin:
Requested by @yugangzhang for CHX.


recipes-tag/mpl-modest-image/meta.yaml
```diff
@@ -0,0 +1,38 @@
+{% set name = "modestimage" %}
+{% set version = "0.2.0" %}
+{% set file_ext = "tar.gz" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  git_url: https://github.com/NSLS-II-CHX/mpl-modest-image
```
tacaswell: why are we looking at their fork rather than upstream?

mrakitin:
The build was successfully tested locally. The recipe was auto-generated with the following command:
```
$ conda skeleton pypi ModestImage
```
I edited the resulted `meta.yml` file to reflect we build from CHX's fork at https://github.com/NSLS-II-CHX/mpl-modest-image, and the tagged version `v0.2.0` (I did it earlier today).

recipes-tag/mpl-modest-image/meta.yaml
mrakitin: See https://github.com/ChrisBeaumont/mpl-modest-image/issues/10 for the request to tag it upstream. No updates since 2017.

tacaswell:
:-1: on adding tags that look anything like the official tags on forked repos (as that is basically a hostile fork!).

I would either just hard-code a SHA in the recipe or tag something like `NSLS-II-v0.2`.

mrakitin:
OK, makes sense to me. I've asked about the new tag [here](https://github.com/ChrisBeaumont/mpl-modest-image/issues/10#issuecomment-458252536).

mrakitin:
Thank you @tacaswell! Good changes.

PR #528, opened 2019-01-28
MNT: pin libpng back on matplotlib

tacaswell:


tacaswell:
I plan to self-merge this before 17:45

PR #530, opened 2019-01-29
Update CHX analysis metapackage

mrakitin:
... to include modestimage (#527).


mrakitin:
Self-merging to make it available for CHX's analysis server.

PR #532, opened 2019-01-29
FIX: Pinning specification had wrong syntax.

danielballan:
Fixing a mistake a merged too quickly in #531. This now pins in exactly
the same way that the analysis metapackage pins it.

### DIFF
--- a/recipes-tag/hdf5-lz4/meta.yaml
+++ b/recipes-tag/hdf5-lz4/meta.yaml
@@ -11,9 +11,9 @@ build:
 
 requirements:
     build:
-        - hdf5 1.10.2
+        - hdf5 =1.10.2
     run:
-        - hdf5 1.10.2
+        - hdf5 =1.10.2
 
 about:
     home: https://github.com/dectris/HDF5Plugin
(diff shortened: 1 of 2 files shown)


recipes-tag/hdf5-lz4/meta.yaml
```diff
@@ -11,9 +11,9 @@ build:
 
 requirements:
     build:
-        - hdf5 1.10.2
+        - hdf5 =1.10.2
```
mrakitin: Should it be a single `=` or double `==`?

mrakitin: https://docs.conda.io/projects/conda-build/en/master/source/package-spec.html#build-version-spec does not have any single `=` as a version matching.

mrakitin:
Update from CHX: the build produced by #531 was successfully installed and works fine (the data from Eiger loaded with no issues).

recipes-tag/hdf5-lz4/meta.yaml
mrakitin: I didn't read the description carefully. I see now that it's the same as in https://github.com/NSLS-II/lightsource2-recipes/blob/c7380dd7aa6f7393581e3a262c417e177feb50ef/recipes-tag/analysis/meta.yaml#L28

PR #542, opened 2019-01-31
BLD: change Docker image name for interactive building

mrakitin:
It's useful when frejya is not healthy.


mrakitin:
I think I'll merge it since it looks like I'm the only person who uses this :)

PR #547, opened 2019-02-04
Add papermill package

mrakitin:
Per @yugangzhang's request for CHX.


mrakitin:
Merging it to build faster and make it available for CHX.

PR #549, opened 2019-02-04
Add missing packages for papermill

mrakitin:
These are the missing dependencies for the papermill package (#547). Tested, they are built fine locally.


mrakitin:
Thanks for merging, @tacaswell!

PR #555, opened 2019-02-04
ENH: build and upload multiple packages

### LINKED ISSUE #551: Take care of multiple packages built with one meta.yaml
The problem is that a package like https://github.com/NSLS-II/lightsource2-recipes/blob/master/recipes-tag/mpi/meta.yaml generates 2 packages:
```
anaconda upload /conda/conda-bld/linux-64/mpi-1.0-mpich.tar.bz2
anaconda upload /conda/conda-bld/linux-64/mpi-1.0-openmpi.tar.bz2
```

which both should be uploaded. However, the current approach in our building system is to upload only one package. This should be resolved. See the relevant code here: https://github.com/NSLS-II/lightsource2-recipes/blob/eea94164d7db8f9293562cad7415c1b1ecaec881/scripts/build.py#L528-L544

mrakitin:
This PR addresses the issue with multiple packages to build within a single metapackage. This is tested on https://anaconda.org/mrakitin/mpi/files, where it built and uploaded 2 packages:
-  `linux-64/mpi-1.0-openmpi.tar.bz2`
- ` linux-64/mpi-1.0-mpich.tar.bz2`

Resolves #551.


mrakitin:
After this is merged, I suggest we rebuild all the packages involved into the original PR #534. It means we will have to first remove the packages from the both public and internal channels, and then start building them again. Objections?

leofang:
LGTM! For rebuilding, how about #525 for matplotlib? Would the steps you suggest break too many things?

licode:
LGTM. I think this should work fine for the case of only one package. self merge when you are ready.

mrakitin:
Thanks for your reviews, @licode and @leofang! Yes, this should be backwards-compatible as `determine_build_name()` is used only in one place, and and the calling code loops over returned packages. Good idea about rebuilding MPL, though it can be addressed via a separate PR. @leofang, do you mind to open one and keep it open until we resolve the MPI problem? Thanks!

PR #563, opened 2019-02-20
Add pyopencl and deps

mrakitin:
This dependency is needed for the [sasview](https://github.com/SasView/sasview) package. If/when my PR https://github.com/SasView/sasview/pull/206 is reviewed/merged, the corresponding recipe of `sasview` can be produced. Meanwhile, the package can be temporary built from the PR branch/commit.


recipes-tag/ocl-icd/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  skip: True  # [not linux]
+
+requirements:
+  build:
+    - pkg-config
+    - libtool
+    - automake
+    - autoconf
+    - {{ compiler('c') }}
+    - ruby
```
mrakitin: This is added by me on top of the [conda-forge's recipe](https://github.com/conda-forge/ocl-icd-feedstock/blob/22625432a0ae85920825dfeb103af9fe7bd6a950/recipe/meta.yaml#L23-L29).

mrakitin:
I spent 3 hours yesterday trying to make it working. I ended up modifying the recipe of `pyopencl` so that it does not depend on `pocl` in the tests to avoid building all the corresponding dependencies. So, eventually, I only had to build 2 packages: `ocl-icd` and `pyopencl`. The resulted `pyopencl` package was built, but on the testing stage of the building there was some dependency resolution issue that the package was incompatible with the current environment. In the debug mode it showed that there was a conflict of python version (3.5), so I'm completely lost there. I assume it can be something to deal with the `host:` section of the recipe, but not sure. As it's an optional dependency of xi-cam and sasview, I'll close this draft PR. We can readdress it later when needed.

leofang:
This is weird. I thought you'd see conflicts due to driver version or whatnot, not due to Python. I never built `pyopencl` myself but I have some experience with `pycuda`, developed in parallel by the same author and supporting CUDA. And building `pycuda` depends on the CUDA driver and runtime library installed in the `run` environment (not necessary same as `host`), so at that time I didn't think it's possible to ask DAMA to support it because everyone has a different version installed. 

A workaround adopted by CuPy is to support all recent versions of CUDA toolkits for the users, see [here](https://docs-cupy.chainer.org/en/stable/install.html#install-cupy):

```shell
(For CUDA 8.0)
$ pip install cupy-cuda80

(For CUDA 9.0)
$ pip install cupy-cuda90

(For CUDA 9.1)
$ pip install cupy-cuda91

(For CUDA 9.2)
$ pip install cupy-cuda92

(For CUDA 10.0)
$ pip install cupy-cuda100
```

This is a rabbit hole that we better not get trapped into.

mrakitin:
Thanks for the insight into the pycuda package, which we may also consider in the future.

leofang:
We've abandoned pycuda and moved to CuPy btw 😄

mrakitin:
Yes, seems to be a better alternative.

PR #565, opened 2019-02-21
Add sasmodels package

mrakitin:
This is a dependency of the sasview package (PR will follow), which should be build first. Tested locally, builds fine. Ready for review/merge.

Also, some minor improvement to the building script, so that it can be executed from any location, not just from `scripts/`.


licode:
LGTM.

mrakitin:
Thank you @licode!

PR #566, opened 2019-02-21
More sasview dependencies

mrakitin:
Discovered 3 more `sasview` dependencies which do not exist in the `defaults` channel:
- `bumps`
- `qt5reactor`
- `xhtml2pdf`


mrakitin:
Thank you @licode!

PR #568, opened 2019-02-22
Add flake8 to collection package

mrakitin:
It was asked a few times how to use flake8. It's probably worth to include it to our deployment collection.


mrakitin:
Thank you, @tacaswell!

PR #574, opened 2019-03-01
Use https links for the mpich recipe

mrakitin:
When I build locally, for some reason my laptop refuses to recognize `http` links. I verified the `https` links work.


leofang:
Not sure why `http` links are not recognized, but we do always prefer `https` links if they're available.

mrakitin:
It happens with my laptop + office network sometimes when I plug/unplug it from the cable a few times during the day. Something weird. Thanks for reviewing/merging!

PR #576, opened 2019-03-02
Mac builds

mrakitin:
Per @awalter-bnl's request, osx64 packages are in our channel now. This required some minor updates to a few recipes:
- where `numpy` was used in the `build` section, it was pinned to the compatible version;
- `nslsii_dev_configuration` package does not seem to be necessary as `databroker` can be imported without that;
- `toolchain` was missing from our recipes - as a result the corresponding package wasn't found. The recipe was copied over from conda-forge.
- bumped a version of `xraylib` to 3.3.0 and added a condition to `build.sh` since one of 5 tests failed on Mac, but worked on Linux.

I tested creating an environment on Mac and successfully installed the `collection` package (which also installed the `analysis` one, so it's pretty much the same we deploy on the floor.) Python 3.6 only packages were built; 3.7 will be there soon. Enjoy!


tschoonj:
Was it the Ruby test that failed for xraylib on macOS?

mrakitin:
@tschoonj, thank you for your question, which led me to the solution!

Yes, I it failed on the Ruby one, here is the output of the test:
```
===========================================
   xraylib 3.3.0: example/test-suite.log
===========================================

# TOTAL: 5
# PASS:  4
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: xrlexample10.sh
=====================

/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require': dlopen(/miniconda3/envs/conda-build/conda-bld/xraylib_1552313251857/work/ruby/.libs/xraylib.bundle, 0x0009): dependent dylib '/miniconda3/envs/conda-build/conda-bld/xraylib_1552313251857/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/libxrl.7.dylib' not found for '/miniconda3/envs/conda-build/conda-bld/xraylib_1552313251857/work/ruby/.libs/xraylib.bundle' - /miniconda3/envs/conda-build/conda-bld/xraylib_1552313251857/work/ruby/.libs/xraylib.bundle (LoadError)
    from /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from ../example/xrlexample10.rb:12:in `<main>'
FAIL xrlexample10.sh (exit status: 1)
```
In 49c0a2d I added the `ruby` package dependency in the build section and reverted the `build.sh` to have `make check` unconditionally, which helped to resolve the issue. Thanks for the hint!

Side question: how did you learn about the problem we had with xraylib for Mac?

tschoonj:
The reason why ruby was failing is [SIP](https://en.wikipedia.org/wiki/System_Integrity_Protection).

Basically all executables in `/usr/bin` no longer take into account the `DYLD_LIBRARY_PATH` variable, which is used by `make check` internally to ensure that the xraylib dylib gets picked up.

You would get the same problem with the python and perl bindings if you would have used the system interpreters.

In the upcoming version of xraylib I have greatly expanded the unit tests, and did in such a way that you will also need to override the `/bin/bash` shell on macOS, for the same reason...

tschoonj:
Oh and I got onto this page because I periodically check Github for issues involving xraylib :)

mrakitin:
@tschoonj, thank you for the explanation! And great job in tracking the downstream consumers of your code. What's the best way to keep in touch if we experience issues (if any) with xraylib?

tschoonj:
Just open an [issue](https://github.com/tschoonj/xraylib).

mrakitin:
Thanks, @tschoonj, we will!

mrakitin:
Thank you @CJ-Wright.

PR #578, opened 2019-03-06
Bump event-model version to 1.8.0.

danielballan:


CJ-Wright:
Does this have a numpy dependency?

danielballan:
Yes. I should have added that. @mrakitin What is the best way to fix this these days?

mrakitin:
@danielballan, do you want to re-tag? If so, we will remove the built/uploaded packages from both public and private conda channels.

CJ-Wright:
You could patch it in the repodata.json

mrakitin:
A new PR with the added dependency of `numpy` and the bumped build number would be sufficient, I suppose.

mrakitin:
> @danielballan, do you want to re-tag? If so, we will remove the built/uploaded packages from both public and private conda channels.

I think I was wrong about re-tagging. The tag is there, it's just the new dependency.

PR #579, opened 2019-03-06
event-model: add numpy dependency

mrakitin:
This is to resolve the conversation in #578. Once merged, I'll remove the built/uploaded packages from anaconda and alexandria, so that they can be rebuilt.

@danielballan, @CJ-Wright, agree with this plan?


mrakitin:
I think as the currently uploaded file is essentially broken (misses an important dependency), it's worth to push it as the same build number.

CJ-Wright:
You might bump the build number.

mrakitin:
I remember @danielballan expressed a concern that the package with the latest build number may not be picked up while installing it for the specific version. It was something regarding anaconda's way to identify the dependencies. Not sure exactly though.

@danielballan, what do you think?

CJ-Wright:
The latest build number should be respected, I think?
You could also move the packages onto an unused label (we move bad things to "broken" at CF)

mrakitin:
I am fine with bumping a build number, thanks @CJ-Wright.

danielballan:
> The latest build number should be respected, I think?

Nope. If conda is given a choice between <set of dependencies> with build number 0 and <larger set of dependencies> with build number 1, it reserves the right to choose the lower build number and the smaller set of dependencies. Build numbers are treated as "flavors" not as sequential number given priority. This has bitten us many, many times.

mrakitin:
Hmm, so the scenario of removing the uploaded packages + having the build number `0` seems the most reliable one in this case. Agree?

danielballan:
Yes.

mrakitin:
Great! Self-merging.

mrakitin:
Ups, I haven't changed the build number...

leofang:
> > The latest build number should be respected, I think?
> 
> Nope. If conda is given a choice between `<set of dependencies>` with build number 0 and `<larger set of dependencies>` with build number 1, it reserves the right to choose the lower build number and the smaller set of dependencies. Build numbers are treated as "flavors" not as sequential number given priority. This has bitten us many, many times.

Is it because we used to use older `conda build`?

mrakitin:
@leofang, I think it's not about the build process itself, it's about installing the resulted packages by conda.

PR #580, opened 2019-03-06
Revert "event-model: add numpy dependency"

mrakitin:
Reverts NSLS-II/lightsource2-recipes#579


mrakitin:
Sorry for the mess. This PR resolves the wrong build #, which should be `0`.
bc95804 takes care of adding `numpy` as a dependency.

PR #585, opened 2019-03-07
matplotlib 3.0.3

mrakitin:
The files were copied over from https://github.com/conda-forge/matplotlib-feedstock/tree/master/recipe. Now we should have 3 packages generated by this recipe:
- `matplotlib-base-3.0.3-py36_0.tar.bz2`
- `matplotlib-3.0.3-py36_0.tar.bz2`
- `mpl_sample_data-3.0.3-py36_0.tar.bz2`

It builds successfully in a local Docker container.


mrakitin:
Thanks for the review, @danielballan!
Sure, @tacaswell, do you have any updates/comments/concerns? It's now a plain copy of the recipe from conda-forge. The only difference is that the `yum_requirements` file was removed as we don't build/install on the RPM-based machines.

PR #586, opened 2019-03-07
SRX packages

mrakitin:
We just realized with @andrewmkiss that we are missing the analysis metapackage for SRX. I pushed the only extra package `pyxrf` manually, but with this PR we will push it automatically in the future.


danielballan:
Is pyxrf now Python 3 compatible? At some point we had to put it in special Python 2 environments.

licode:
py36 now. LGTM.

mrakitin:
@danielballan, yes, it was successfully installed in our standard analysis-2019-1.2-srx environment. Great work on making it compatible with Python 3.6+, @licode.

licode:
Thanks for adding this, @mrakitin . I should provide more update on pyxrf, so we can all stay at the same page without getting confused.

mrakitin:
Cool, thanks for merging, @licode.

PR #590, opened 2019-03-08
Fix suitcase-json-metadata and suitcase-specfile sha256

mrakitin:
Overlooked while reviewed #588.

Attn @danielballan @awalter-bnl


danielballan:
Ah, thanks. I'm still learning my way around the improved system we have for these things.

mrakitin:
Me too :(
I hope soon we will develop a better system with in-advance tests of the recipes.

PR #603, opened 2019-03-13
event-model: make it 3.6+

mrakitin:
To mirror the changes in https://github.com/conda-forge/event-model-feedstock/pull/10 as event-model 1.8.2 is Python 3.6+ (https://github.com/NSLS-II/event-model/pull/51). See discussion in Slack with @CJ-Wright for details.

PR #606, opened 2019-03-13
event-model: more fixes

mrakitin:
This PR resolves the issue observed in #603: we don't limit the Python's version in the `build` section.


CJ-Wright:
We usually do it in both but we also don't have a build section (we have a host section when compilers aren't involved)

recipes-tag/event-model/meta.yaml
```diff
@@ -9,13 +9,13 @@ source:
   git_rev: v{{ version }}
 
 build:
-  number: 1
+  number: 2
   script: python setup.py install --single-version-externally-managed --record=record.txt
   skip: True  # [py2k]
 
 requirements:
   build:
```
CJ-Wright:
```suggestion
  host:
```

CJ-Wright:
Ah I see the issue this is not a noarch recipe (you can just move the selector to ``py>=36``)

mrakitin:
Yeah, it's already arch...

> (you can just move the selector to `py>=36`)
what do you mean by that?

recipes-tag/event-model/meta.yaml
mrakitin: I hope it won't break anything.

danielballan:
Thanks @CJ-Wright!

recipes-tag/event-model/meta.yaml
```diff
@@ -9,13 +9,13 @@ source:
   git_rev: v{{ version }}
 
 build:
-  number: 1
+  number: 2
   script: python setup.py install --single-version-externally-managed --record=record.txt
   skip: True  # [py2k]
```
CJ-Wright:
```suggestion
  skip: True  # [py>=3.6]
```
From https://docs.conda.io/projects/conda-build/en/latest/source/resources/define-metadata.html#preprocessing-selectors

recipes-tag/event-model/meta.yaml
```diff
@@ -9,18 +9,18 @@ source:
   git_rev: v{{ version }}
 
 build:
-  number: 1
+  number: 2
   script: python setup.py install --single-version-externally-managed --record=record.txt
-  skip: True  # [py2k]
+  skip: True  # [py>=3.6]
```
mrakitin: Are we sure we skip the newer versions? I think it should be `py<3.6`...

recipes-tag/event-model/meta.yaml
mrakitin: Thanks for the link! I think the condition should be opposite. See my comment below.

recipes-tag/event-model/meta.yaml
mrakitin: OK, looks good now.

mrakitin:
@CJ-Wright, thanks for a few commits, feel free to merge whenever you are ready.

mrakitin:
OK, self-merge to have it built sooner.

CJ-Wright:
sorry

PR #608, opened 2019-03-14
Add grid-strategy

mrakitin:
Per @tacaswell's request.


recipes-tag/grid-strategy/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+requirements:
+  build:
+    - python
+    - setuptools
+  run:
+    - python
```
tacaswell: This requires Matplotlib as well

mrakitin: I see. It's not in `setup.py` for some reason.

mrakitin: Done.

danielballan: This comment made me curious. It's [in their `setup.cfg`](https://github.com/matplotlib/grid-strategy/blob/db12cdf788f6ddca7cec231753b2014437942db3/setup.cfg#L26-L28) instead. I have not seen that approach before. I suppose the advantage is that INI is more descriptive and easier to externally parse than Python.

mrakitin: Yeah, interesting, thanks for digging into the details, @danielballan! I also see there is a `numpy` dependency, and it's 3.6+ only. I'll make the corresponding changes to the recipe. Stay tuned.

tacaswell:
Thanks for taking care of this @mrakitin !

PR #609, opened 2019-03-14
mayavi: do not skip 3.7+

mrakitin:
Attn @CJ-Wright, this should fix the issue with mayavi build. Now we have all necessary tools (`Xvfb` and `killall`). It builds fine locally. Let's try with freyja.


mrakitin:
Thanks!

PR #610, opened 2019-03-14
Analysis 2019-2.0

mrakitin:
No suitcases this time. Per our discussion with @danielballan in Slack, the old `suitcase` is already explicitly installed via ansible's `extra_packages`. The new `suitcase-{tiff,csv,jsonl,json-metadata}` packages will be put manually into the extra-packages if beamlines don't use the old `suitcase` for HDF5 exports.

Also, I put the minimum versions for bluesky software stack that we just tagged for this deployment to avoid surprises.


mrakitin:
Thanks for the review. Self-merging.

PR #611, opened 2019-03-14
Grid strategy skip py older 36

mrakitin:
Overlooked in #608. This should be the same as we did for event-model yesterday (see #606).


mrakitin:
Thank you Andrew!

PR #616, opened 2019-03-15
isstools v3.0.0 and xas v0.0.1 packages

mrakitin:
Tagged the version of ISS packages. Need to have them at the beamline during the deployment 19-2.


mrakitin:
`xas` was tested to build locally, worked fine. `isstools` depends on that, so I wasn't able to test it via a separate Docker container. Hopefully, it will be built fine. Merging to check it.

PR #617, opened 2019-03-15
XFM collection metapackage

mrakitin:
To have a metapackage for https://github.com/NSLS-II/playbooks/pull/131 rather than using `hxntools` directly in `extra_packages`.


mrakitin:
Self-merging.

PR #619, opened 2019-03-18
ENH: use the "a" mode to append to a file rather than overwrite it

mrakitin:
This is to fix the issue with the overwritten log file during the mirroring process. In particular, this line
https://github.com/NSLS-II/lightsource2-recipes/blob/a5f2de107e51f8813afb0e487466d7e3cdae9eca/scripts/mirror-to-nsls2-local-folder.sh#L23
overwrites the log produced by 
https://github.com/NSLS-II/lightsource2-recipes/blob/a5f2de107e51f8813afb0e487466d7e3cdae9eca/scripts/mirror-to-nsls2-local-folder.sh#L21


mrakitin:
To make it easier to find in the future, I like this SO answer: https://stackoverflow.com/a/30566011/4143531. Picture from there:
image

mrakitin:
OK, self-merging to have better logs from the mirroring part.

PR #624, opened 2019-03-20
Add a stricter pin of analysis in the collection package

mrakitin:
I hope this time it will work properly.


mrakitin:
It builds fine locally. Will merge to proceed.

Issue #625, opened 2019-04-02
Sort out mayavi build issue


mrakitin:
It was attempted to build via https://github.com/NSLS-II/lightsource2-recipes/pull/604, but we eventually gave up in https://github.com/NSLS-II/lightsource2-recipes/pull/612.

mrakitin:
No, it still fails with:
```
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {"python[version='>=3.6,<3.7.0a0']", 'python=3.7'}
```

CJ-Wright:
where is that coming from?

mrakitin:
A more detailed traceback:
```py
No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11
WARNING:conda_build.metadata:No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.11
Adding in variants from internal_defaults
Adding in variants from argument_variants
INFO:conda_build.variants:Adding in variants from internal_defaults
INFO:conda_build.variants:Adding in variants from argument_variants
Collecting package metadata: ...working... done
Solving environment: ...working... done
Collecting package metadata: ...working... done
Solving environment: ...working... failed
Traceback (most recent call last):
  File "/conda/tmp_envs/55fc582f7dd397d3a5f0/lib/python3.7/site-packages/conda_build/environ.py", line 753, in get_install_actions
    actions = install_actions(prefix, index, specs, force=True)
  File "/conda/tmp_envs/55fc582f7dd397d3a5f0/lib/python3.7/site-packages/conda/common/io.py", line 85, in decorated
    return f(*args, **kwds)
  File "/conda/tmp_envs/55fc582f7dd397d3a5f0/lib/python3.7/site-packages/conda/plan.py", line 473, in install_actions
    txn = solver.solve_for_transaction(prune=prune, ignore_pinned=not pinned)
  File "/conda/tmp_envs/55fc582f7dd397d3a5f0/lib/python3.7/site-packages/conda/core/solve.py", line 107, in solve_for_transaction
    force_remove, force_reinstall)
  File "/conda/tmp_envs/55fc582f7dd397d3a5f0/lib/python3.7/site-packages/conda/core/solve.py", line 145, in solve_for_diff
    force_remove)
[...]

PR #627, opened 2019-04-02
use new vtk

CJ-Wright:


mrakitin:
It was built successfully: https://anaconda.org/lightsource2-tag/mayavi/files.

Closes #625.

PR #639, opened 2019-04-04
Add missing packages to the ISS analysis metapackage

mrakitin:
2 packages were missing during the latest push:
- pysmbc
- periodictable (probably it was never rebuilt with it)

PR #640, opened 2019-04-04
Use the updated analysis metapackage; remove unused packages

mrakitin:
Only local processing is used for now, that's why we don't need the `lightflow` and `kafka-python` packages for now.

#639 should be built first, that's why separate packages.


mrakitin:
It's good to go now as the packages from #641 and #639 have been built already.

PR #641, opened 2019-04-04
Add pysmbc dep to isstools

mrakitin:
It should be built prior to #639 and #640.


mrakitin:
It's reverting the package mistakenly removed via #616.

mrakitin:
Self-merging to have it built at 2:45 pm.

PR #644, opened 2019-04-08
Correct qastools conda package name

mrakitin:
I used a wrong name of the `qastools` conda package in #642. This PR resolves it (will remove the wrong packages from anaconda.org and alexandria manually).


mrakitin:
Self-merging.

PR #649, opened 2019-04-09
rebuild fabio for numpy 1.15

CJ-Wright:


mrakitin:
We have the latest build (29 days ago) with the same config here: https://anaconda.org/lightsource2-tag/fabio/files. Not sure it will help to resolve the issue.

PR #648, opened 2019-04-09
ENH: add tests to qastools recipe

mrakitin:
The motivation is an hour spent for tracking down [this](https://github.com/NSLS-II-QAS/isstools/commit/78524b5c49541f7c0baaa97ed95293e46d0110bb) missing `__init__.py`.

Besides that I suspect that our Docker image does not perform testing, as I see those messages when built locally (without pushing):
```
...
Fixing permissions
Compressing to /tmp/tmpediah7c1/qastools-3.0.1qas-py36_1.tar.bz2
Importing conda-verify failed.  Please be sure to test your packages.  conda install conda-verify to make this message go away.
WARNING:conda_build.build:Importing conda-verify failed.  Please be sure to test your packages.  conda install conda-verify to make this message go away.
...
```

However, when I run a build directly in the container with `conda build .`, I get this:
```
...
Fixing permissions
Compressing to /tmp/tmpian4ismf/qastools-3.0.1qas-py36_1.tar.bz2
TEST START: /conda/conda-bld/linux-64/qastools-3.0.1qas-py36_1.tar.bz2
Adding in variants from /tmp/tmpx29gahe2/info/recipe/conda_build_config.yaml
INFO:conda_build.variants:Adding in variants from /tmp/tmpx29gahe2/info/recipe/conda_build_config.yaml
Renaming work directory,  /conda/conda-bld/qastools_1554843658637/work  to  /conda/conda-bld/qastools_1554843658637/work_moved_qastools-3.0.1qas-py36_1_linux-64
Solving environment: ...working... done
...
```

PR #650, opened 2019-04-12
qastools update

mrakitin:
This is to separate `qastools` from #648 (which is essentially a building system fix now).


mrakitin:
Self-merging to get ready for the QAS visit.

PR #653, opened 2019-04-16
CSX: analysis metapackage update

mrakitin:
This package should be built first (before the collection one) as the collection needs this analysis metapackage.


mrakitin:
Thanks @gwbischof!

PR #658, opened 2019-04-17
Added jupyterlab and nodejs to FMX meta package

gwbischof:


mrakitin:
Looks good to me. This is to run jupyterlab for collection at FMX and enable extensions with nodejs.

PR #662, opened 2019-04-22
xpd April 22nd 2019

CJ-Wright:


CJ-Wright:
I think this is mostly ok, self merging.

PR #663, opened 2019-04-22
remove mayavi again

CJ-Wright:


mrakitin:
Is `mayavi` still causing issues?

CJ-Wright:
Yes, it caused an unsatisfiable dependency error for conda install

PR #661, opened 2019-04-19
Fix tifffile recipe

mrakitin:
The following message appears when `bsui` starts:
```py
/opt/conda_envs/collection-2019-2.0-csx/lib/python3.6/site-packages/tifffile/tifffile.py:298: UserWarning: ImportError: No module named '_tifffile'. Loading of some compressed images will be very slow. Tifffile.c can be obtained at http://www.lfd.uci.edu/~gohlke/
  "ImportError: No module named '_tifffile'. "
```
According to @CJ-Wright, `{{ pin_compatible('numpy') }}` can only be in the `run` or `test` section of the recipe, but not in the `built` one.

PR #667, opened 2019-04-30
Fix xraylib package

mrakitin:
Hopefully it should resolve the package the same way as #661 (`tifffile`) and https://github.com/NSLS-II/lightsource2-recipes/pull/656 (`fabio`) after breaking it via #576 (Mac builds).

Builds fine locally.


danielballan:
Thank you. I meant to take care of this while you were away.

mrakitin:
Thanks for merging @danielballan. No worries, hopefully we can test it somewhere soon. I think we can push silently to BMM machine(s) and check if it imports without loading the profile in pure Python or default IPython session. What do you think?

mrakitin:
For the record, the `xraylib` is a package we deploy everywhere on the floor with the `analysis` meta-package https://github.com/NSLS-II/lightsource2-recipes/blob/53a4825ae55e2bdd84d2b5d9f59306d061681776/recipes-tag/analysis/meta.yaml#L50
(which is a dependency of the `collection` meta-package as well), via scikit-beam's dependency
https://github.com/NSLS-II/lightsource2-recipes/blob/53a4825ae55e2bdd84d2b5d9f59306d061681776/recipes-tag/scikit-beam/meta.yaml#L31

I tested the built package on xf06bm-ws2 under collection env by importing `xraylib` in ipython before and after the push of the new package, and the new version fixes the problem. I pushed it to all the BMM machines to both analysis and collection environments.

@bruceravel, let us know if anything else is needed.

bruceravel:
Will do.  It may take me a while to get to it.  I am pretty packed with users for the near future, so I'll continue using the collection from the last cycle for a while longer.

mrakitin:
@bruceravel, thanks for letting us know and sorry for the broken package. The lesson is learned - we will do better testing before pushing to the floor.

PR #669, opened 2019-05-02
Bump caoroto to 0.3.4.

danielballan:
Merge #668 first and let that build.


mrakitin:
After #668 is merged, this PR is conflicting. As requested, I'll wait until after v0.3.3 from #668 is built anyway.

PR #682, opened 2019-06-07
MNT: temporarily set the build window to 3 years

tacaswell:
The goal is to force everything that does not have a py37 build
to rebuild it's self.  Once that is done, we can go back to 3 days.


mrakitin:
Can we wait a bit with this? As our building system is still lacking the testing part, we may have some broken builds. I will try to solve it tomorrow.

mrakitin:
Agreed over Slack communication with @tacaswell to merge it to see if it works, and test manually next week.

PR #683, opened 2019-06-07
TES: add PyXRF dependency

mrakitin:
xref https://controlsweb.nsls2.bnl.gov/trac/ticket/3531.


ke-zhang-rd:
LGTM. Any comments or concerns?

PR #684, opened 2019-06-07
BLD: Update collection package to version 2019C3.0

ke-zhang-rd:
Add two kafka package


mrakitin:
For the confluent-kafka, copy over the following recipe, and place it to the same PR: https://github.com/conda-forge/python-confluent-kafka-feedstock/tree/master/recipe. Thanks!

PR #685, opened 2019-06-07
BLD: meta file for pykafka

ke-zhang-rd:
https://github.com/Parsely/pykafka/releases

2.8.0 latest release for pykafka

### DIFF
--- a/recipes-tag/pykafka/meta.yaml
+++ b/recipes-tag/pykafka/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "pykafka" %}
+{% set version = "2.8.0" %}
+{% set sha256 = "f0bbd394ae6970042a587c99fe4dc0966e67787249d963d4ce2f810dc9490577" %}
+
+package:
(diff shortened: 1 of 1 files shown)


recipes-tag/pykafka/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - {{ compiler('c') }}
+  host:
+    - python
+    - setuptools
+    - librdkafka              # [not win]
+  run:
+    - python
+    - setuptools
+    - six >=1.5
+    - kazoo >=2.5.0
```
mrakitin: This package does not exist in the `defaults` channel.

recipes-tag/pykafka/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - tabulate
+    - gevent
+    - librdkafka              # [not win]
```
mrakitin: This should be copied over from conda-forge feedstock repo as well.

PR #686, opened 2019-06-07
Add suitcase-msgpack recipe.

danielballan:


mrakitin:
This package was never built. The problem was with the hash value. I'm going to create a PR which resolves it.

PR #688, opened 2019-06-07
BLD: Add librdkafka

ke-zhang-rd:
Edited by @mrakitin: it's a dependency of both pykafka (#685) and confluence-kafka (#687).

### DIFF
--- a/recipes-tag/librdkafka/meta.yaml
+++ b/recipes-tag/librdkafka/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "librdkafka" %}
 {% set version = "0.11.6" %}
-{% set sha256 = "9c0afb8b53779d968225edf1e79da48a162895ad557900f75e7978f65e642032" %}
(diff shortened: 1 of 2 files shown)


recipes-tag/librdkafka/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "librdkafka" %}
+{% set version = "0.11.6" %}
```
mrakitin: I think we can use v1.0.1 here. The corresponding conda-forge feedstock PR is passing: https://github.com/conda-forge/librdkafka-feedstock/pull/33.
```suggestion
{% set version = "1.0.1" %}
```

recipes-tag/librdkafka/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+{% set sha256 = "9c0afb8b53779d968225edf1e79da48a162895ad557900f75e7978f65e642032" %}
```
mrakitin:
```suggestion
{% set sha256 = "b2a2defa77c0ef8c508739022a197886e0644bd7bf6179de1b68bdffb02b3550" %}
```

PR #690, opened 2019-06-07
BLD: update analysis for 2019C3

ke-zhang-rd:
We need bluesky and ophyd tagged as shown in this PR first


mrakitin:
Note, it should be merged _before_ #684, as it's a dependency of the collection metapackage.

PR #694, opened 2019-06-10
BLD: add lxml and update cycle

ke-zhang-rd:
Based on Masa requests


recipes-tag/11-bm-cms-collection/meta.yaml
```diff
@@ -12,3 +12,4 @@ build:
 requirements:
   run:
     - natsort
+    - lxml
```
mrakitin: I'd recommend to sort the dependencies alphabetically.

PR #699, opened 2019-06-11
16id Package needed for 2019C3.0

ke-zhang-rd:
Add, all exist in lightsource2-recipe
`py4xs`
`lmfit`

### DIFF
--- a/recipes-tag/16-id-lix-analysis/meta.yaml
+++ b/recipes-tag/16-id-lix-analysis/meta.yaml
@@ -12,3 +12,6 @@ build:
 requirements:
   run:
     - fabio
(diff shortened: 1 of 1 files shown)


recipes-tag/16-id-lix-analysis/meta.yaml
```diff
@@ -12,3 +12,6 @@ build:
 requirements:
   run:
     - fabio
+    - py4xs
+    - lmfit >=0.9.9
+    - peakutils
```
mrakitin: Peakutils is a general dependency of the `analysis` package and does not need to be included.

PR #701, opened 2019-06-11
BLD: pull changes from conda-forge epics-base

tacaswell:


tacaswell:
It can not hurt to carry them.

klauer:
I think this PR caused pyepics CI to fail:
https://github.com/pyepics/pyepics/pull/175
https://travis-ci.org/pyepics/pyepics/jobs/548269555

Either `caget` et al are missing, or the PATH settings are wrong...

PR #702, opened 2019-06-11
BLD: Init scikit-build

ke-zhang-rd:
This is needed for tomopy 1.5.0
meta file is from here
https://github.com/conda-forge/scikit-build-feedstock/blob/master/recipe/meta.yaml


tacaswell:
this needs to be down one level (in a scikit-build folder).

PR #704, opened 2019-06-11
BLD: time stamp and build number update

ke-zhang-rd:
Update time stamp to 2019C3.0 and code style. Validate there is a space before "0" of version.

PR #707, opened 2019-06-11
BLD: Update collection and analysis versiton to 0rc1

ke-zhang-rd:
Use 0rc1 for build env now


ke-zhang-rd:
Work move to here
https://github.com/NSLS-II/lightsource2-recipes/pull/704

PR #710, opened 2019-06-13
BLD: Add lmfit to 17id

ke-zhang-rd:


recipes-tag/17-id-fmx-collection/meta.yaml
```diff
@@ -13,3 +13,4 @@ requirements:
   run:
     - jupyterlab
     - nodejs
+    - lmfit
```
mrakitin: Please sort the dependencies alphabetically. It will make it easier to navigate through the dependencies later.

ke-zhang-rd:
This will merge to other PR

PR #724, opened 2019-06-19
SRW update: use ochubar/SRW as a main repo

mrakitin:
Tested locally, builds fine.

attn @dhidas


mrakitin:
Going to self-merge, needed for Friday's activity with Dean.

PR #726, opened 2019-06-23
Bump nsls2ptycho (ptychography) version

leofang:
Additional changes:
- Support Python 3.7
- Avoid buggy Numba (numba/numba#4164)


leofang:
Merge at will as this PR only triggers the daemon @leofang, not the cron :wink:

PR #727, opened 2019-06-24
bufix tag of xpdAn

CJ-Wright:


CJ-Wright:
@NSLS-II/dama I think this is ready to go.

PR #731, opened 2019-06-24
fix url for xpdan

CJ-Wright:


recipes-tag/xpdan/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
     - pyyaml
     - scikit-beam
     - pyfai
+    - pyxdameraulevenshtein
```
mrakitin: @CJ-Wright, why was this reverted? I think we decided to remove it in the earlier PRs (#729).

CJ-Wright: Sorry I completely missed that

mrakitin: No worries, we tracked this down and fixed via #737. The `xpd*` packages are compatible with the `*-2019-3.0` envs.

CJ-Wright: Thank you so much!

PR #732, opened 2019-06-25
CLN: clean scripts folder from unused files and scripts

mrakitin:
Removing a few outdated scripts which we are not going to use.


mrakitin:
@tacaswell, any opinion about these old files?

PR #733, opened 2019-06-25
BLD: fix building scripts to match the new Docker image v0.1.0

mrakitin:
I tagged the new Docker image produced by https://github.com/NSLS-II/debian-with-miniconda/pull/18 to `v0.1.0`, so that we can control the version/contents of the image we use for building.

This PR also removes the duplicate actions which are a part of our [Dockerfile](https://github.com/NSLS-II/debian-with-miniconda/blob/master/Dockerfile).

Also, as we started using a regular user inside the containers for builds, there are a few corrections to the file paths relevant to the changes in Docker.


mrakitin:
`lmfit-c` and `pycentroids` were built successfully based on this branch, I think it's good to go.

mrakitin:
It's going to be used for production from now on. Self-merging as it worked fine.

PR #735, opened 2019-06-25
Added tests to pycentroids

stuwilkins:
I added the tests to the conda build script. With this not being a pure python module I like going past the import stage to test to make sure that the cxx modules were compiled ok.

### DIFF
--- a/recipes-dev/pycentroids/meta.yaml
+++ b/recipes-dev/pycentroids/meta.yaml
@@ -15,7 +15,9 @@ requirements:
     - setuptools
     - python
     - numpy x.x
+    - scipy
     - pytest
+    - pytest-runner
     - cmake
     - pandas
     - lmfit-c
@@ -28,12 +30,15 @@ requirements:
     - pandas
     - python
     - numpy x.x
+    - scipy
 
 test:
   requires:
     - pytest
   imports:
     - pycentroids
+  commands:
+    - pytest --pyargs pycentroids.tests
 
 about:
   home: https://github.com/NSLS-II/centroids
--- a/recipes-tag/lmfit-c/meta.yaml
+++ b/recipes-tag/lmfit-c/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: lmfit-c
-  version: {{ environ['GIT_DESCRIBE_TAG'][1:] }}.post{{ environ['GIT_DESCRIBE_NUMBER'] }}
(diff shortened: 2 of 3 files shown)


mrakitin:
It's a shame to confess that our building system does not perform tests during the building process (was broken some time ago, still to be fixed). I will have a look at a possible solution today before it gets merged.

stuwilkins:
@mrakitin @leofang I hard coded the version in order to address #736

I still would like these tests, even if we don't build to them because they get run when you manually build

recipes-tag/lmfit-c/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: lmfit-c
-  version: {{ environ['GIT_DESCRIBE_TAG'][1:] }}.post{{ environ['GIT_DESCRIBE_NUMBER'] }}
+  version: "8.2.2.post27"
```
mrakitin: Thanks! I am not sure if we need the build number to be reset, let's try as is.

recipes-dev/pycentroids/meta.yaml
```diff
@@ -15,7 +15,9 @@ requirements:
     - setuptools
     - python
     - numpy x.x
+    - scipy
```
mrakitin: As this recipe was change, we will have to increment the build number above, otherwise the package won't be built.

mrakitin: By the way, why it was set to 10? We usually put `0` as a default. Conda forge may have `1000` to hint which compiler to use, iirc. Right, @CJ-Wright?

stuwilkins: @mrakitin I don't think we need to bump the build number because I bumped the version number?

stuwilkins: I can reset to zero when I do the next version bump, I entered 10 in testing

tacaswell: This is in the dev version which we are not auto-building and should not be manually bumped ;)

stuwilkins: So this is a mute point?

mrakitin: Ah, right, didn't pay attention it was a dev recipe. Probably no point to update it as we don't support it *right now*, but may do in the future.

PR #738, opened 2019-06-26
Fix swift-t recipe

mrakitin:
This should fix the recipe of `swift-t`; tested locally in Docker for both Python 3.6 and 3.7 -- builds successfully.

I added the `sha256` hash value, so that next time the archive with the same version is overwritten, we will know about it.

attn @tacaswell, @ke-zhang-rd


mrakitin:
Self-merged to have it built at 10:00 am on freyja (the crontab job is resumed).

PR #743, opened 2019-07-02
Bump xraylarch version to 0.9.45

mrakitin:
Per @elistavitski's request.

**NOTE**: anaconda.org seems to be down, so it's not possible to test at the moment. Revisit/test once it's back online before merging.


recipes-tag/xraylarch/meta.yaml
```diff
@@ -1,52 +1,56 @@
...
+  patches:
+    - 0001-MNT-remove-the-desktop-shortcut-logic.patch
+    - 0002-MNT-remove-the-feff-wrappers-and-binaries.patch
 build:
-    number: 2
-    script: python setup.py install  # --single-version-externally-managed --record=record.txt
-    skip: True  # [py2k]
+  number: 0
+  script: |
+    # we do not have a license to distribute build FEFF binaries
```
mrakitin:
```suggestion
    # We do not have a license to distribute built FEFF binaries
```

recipes-tag/xraylarch/meta.yaml
```diff
@@ -1,52 +1,56 @@
...
+    # We may have one to deploy it to the floor, but definitly not to make them public
```
mrakitin:
```suggestion
    # We may have one to deploy it to the floor, but definitely not to make them public
```

PR #744, opened 2019-07-02
Update the dependencies of ISS metapackages with pyzbar

mrakitin:
This is WIP as we also need a newer version of xraylarch (#743), which requires more work (new dependencies, etc.)


recipes-tag/08-id-iss-analysis/meta.yaml
```diff
@@ -16,4 +16,5 @@ requirements:
     - netcdf4
     - numexpr
     - periodictable
-    - xraylarch
+    - pyzbar
+    - xraylarch >=0.9.45
```
mrakitin: This is to use the updated `xraylarch` package from https://github.com/NSLS-II/lightsource2-recipes/pull/743.

mrakitin:
A newer version of `xraylarch` was built before via #743. This is ready to go in.

PR #745, opened 2019-07-03
Bump version to 1.8

stuwilkins:
Updated version to 1.8. Attn @ke-zhang-rd and @mrakitin


mrakitin:
This version was never built. Sorry we didn't have any automation in place that time. See #818 for the follow-up discussion.

PR #758, opened 2019-07-30
Bump databroker to v0.13.1.

danielballan:


mrakitin:
Let's build it first, so we have this version available in our conda channel.

PR #762, opened 2019-08-05
Do prerequisites for bluesky-darkframes recipe.

danielballan:


recipes-tag/frozendict/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  run:
+    - python
+
+test:
+  imports:
+    - frozendict
+
+about:
+  home: https://github.com/slezica/python-frozendict
```
mrakitin: This package seems to be an implementation of [this SO answer](https://stackoverflow.com/a/28452633), but with a few additions. Good to have it as a standalone library anyway.

PR #765, opened 2019-08-06
Update QAS metapackage with suitcase-tiff and hxntools

mrakitin:
Observed during a BL visit with @tacaswell and @lcdesilva.


mrakitin:
Thank you, @awalter-bnl!

PR #768, opened 2019-08-12
bump nsls2ptycho version

leofang:


leofang:
Self-merging as it does no harm.

PR #770, opened 2019-08-16
Bump suitcase-utils version

mrakitin:
Based on the latest tag by @jklynch: https://pypi.org/project/suitcase-utils/0.5.0/#files.


mrakitin:
Thanks, @danielballan!

PR #771, opened 2019-08-16
Make suitcase-utils noarch

mrakitin:
Tested via https://github.com/nsls-ii-forge/suitcase-utils-feedstock/pull/1.


mrakitin:
Thank you, @tacaswell!

PR #775, opened 2019-09-03
scikit-beam v0.0.18

mrakitin:
A mirror update based on https://github.com/nsls-ii-forge/scikit-beam-feedstock/pull/4.


mrakitin:
Thanks.

PR #783, opened 2019-09-04
Amostra v1.0.0

mrakitin:
Based on https://github.com/NSLS-II/amostra/releases/tag/v1.0.0.

A mirror update from https://github.com/nsls-ii-forge/amostra-feedstock/pull/2.


mrakitin:
This seems to be a single recipe update missing from the recent updates in this list (general packages, not BL-specific): https://anaconda.org/nsls2forge/repo?sort=time.modified&sort_order=desc.

image

PR #791, opened 2019-09-19
ophyd v1.4.0rc2

mrakitin:
A mirror update of https://github.com/nsls-ii-forge/ophyd-feedstock/pull/6.

PR #787, opened 2019-09-17
Add hxntools to the SRX analysis metapackage

mrakitin:
@andrewmkiss requested the `hxntools` package to be added to the SRX analysis metapackage as it has some critical handlers useful for extracting of the experimental data for analysis in Jupyter.

The PR adds the requested package.

PR #793, opened 2019-09-20
CI: fix git diff command

mrakitin:
The problem with the previous implementation was that for the ~old branches it also took into account the changes which appeared in master from merging the other PRs. E.g., in https://travis-ci.org/NSLS-II/lightsource2-recipes/jobs/587225596 (corresponding to #787) there was a mixture of files corresponding to that PR and the later merged #791:

```
...
Changed files:
recipes-tag/05-id-srx-analysis/meta.yaml
recipes-tag/ophyd/meta.yaml
```

This PR resolves the potential issue and uses the 3-dot notation to extract only the files changed since the merge-base point -- thanks to the absolutely awesome explanation in https://stackoverflow.com/a/7256391/4143531!


.travis.yml
```diff
@@ -4,6 +4,14 @@ services:
   - docker
 sudo: false
 
+env:
+  global:
+    - REMOTE_NAME: upstream
```
danielballan: I like the configurability here in general. It improves the readability of what follows, as well as of course making it more flexible. As a small point, is it necessary to make `REMOTE_NAME` configurable? Setting as an internal constant would shorten the list of “knobs” that one has to understand to use this. Just a thought. If there is a use case for a custom `REMOTE_NAME` that I’m not seeing them of course it should remain.

mrakitin: It was used in two places, so I decided to name it once there. I don't mind to just hard-code it as we should never collide with the existing names on travis.

mrakitin: Addressed via https://github.com/NSLS-II/lightsource2-recipes/pull/793/commits/9b9cb32d47d8928b776a75c2e6bfb5cdc6ad69c3.

PR #794, opened 2019-09-21
Add more info about the env and use TRAVIS_COMMIT_RANGE instead of HEAD

mrakitin:
This should resolve the issue with missing files from the list of touched files for a PR, reported in https://github.com/NSLS-II/lightsource2-recipes/pull/648#issuecomment-533590676.


mrakitin:
Looks like that's what we need.

PR #804, opened 2019-10-17
pyCHX v4.0.6

mrakitin:
Maintenance update based on the new tag https://github.com/NSLS-II-CHX/pyCHX/releases/tag/v4.0.6.

[nsls-ii-forge](https://github.com/nsls-ii-forge) feedstock won't be created for this package.


mrakitin:
Weird the problem with https://travis-ci.org/NSLS-II/lightsource2-recipes/jobs/599382770#L668-L672 wasn't caught. Documented it in the main repo: https://github.com/NSLS-II-CHX/pyCHX/issues/27. It has already been a problem for more than a year, and nobody complained. I think it's OK to go.

PR #805, opened 2019-10-18
Move pychx to recipes-private

mrakitin:
The source repo has been made private, the corresponding recipe needs to be private (= available for internal builds only) as well.

I modified the URL of the sources and removed `pychx` from the list of dependencies of the `11-id-chx-analysis` metapackage. It's not used anywhere else (including `profile_collection`).


leofang:
BTW, is it necessary to remove old builds of `11-id-chx-analysis` from Anaconda after this is merged? I suspect it would attempt to install pyCHX, which now should be already removed?

mrakitin:
Thanks for the review, @leofang.

Regarding the older versions of the package - I don't think anybody else is interested or is using that _meta_ package besides CHX, so I think it's OK to keep the older versions of the metapackage, as it's under our control, and we know what's going on.

PR #816, opened 2019-11-18
pychx v4.0.7

mrakitin:
Has the fix from https://github.com/NSLS-II/pyCHX/pull/30.
Based on the tag: https://github.com/NSLS-II/pyCHX/releases/tag/v4.0.7.


mrakitin:
Tested locally in a Docker container with the passed `GH_TOKEN`, works fine. Going to build it internally and put it to our internal channel.

attn @leofang, please have a look how the version can be parametrized.

PR #825, opened 2020-02-14
Updated hash for lmfit-c

stuwilkins:
@mrakitin This is the updated version of `lmfit-c` with the correct hash....


mrakitin:
Thanks, @stuwilkins!!!

PR #829, opened 2020-03-23
Create package 08-bm-tes-analysis

dmgav:


danielballan:
For the upcoming deployment, we plan to put the union of all beamlines' requirements in _one_ metapackage (or the smallest number we can manage, if we encounter mutually incompatible requirements) instead of creating a separate metapackage per beamline. Thus, I think this would never be used.

This work is being done by @jklynch and @mrakitin .

dmgav:
Actually, I was instructed to create this package: https://github.com/NSLS-II/playbooks/pull/346
I would the environment 2019-3.0 to be deployed on the server so that I could demonstrate that PyXRF is running on it. Can it be done differently?

mrakitin:
@dmgav, sorry for the conflicting instructions. For the new conda environments, which will go to the experimental floor during the upcoming shutdown, we indeed plan to have a single analysis environment with all packages added together.

However, since you are planning to use the existing ansible infrastructure, which still uses separate metapackages for each beamline, I think it's OK to have this beamline-specific metapackage for the `2019-3.0` deployment on the new machine.

mrakitin:
@danielballan, do you agree with that plan?

jklynch:
The [nsls2 analysis metapackage](https://github.com/nsls-ii-forge/nsls2-analysis-feedstock) currently lists hxntools and pyxrf:
```
    - hxntools
    - pyxrf >=0.0.9
```
I just copied these from SRX (I think). Should I specify different versions?

dmgav:
The latest versions: PyXRF v0.0.26, scikit-beam v0.0.21. I think we should  specify those versions. It makes no sense to install earlier versions. The problem with PyXRF is that I update it very often, so I will probably still have to install both PyXRF and scikit-beam from the repositories. But it helps to have all other packages up to date.

mrakitin:
@jklynch, pyxrf v0.0.26 seems to be the latest in https://anaconda.org/nsls2forge/pyxrf/files for now.

The synchronization of versions in this, `lightsource2-tag` channel stopped after pyxrf v0.0.11. I think we need to catch up with the latest versions of pyxrf/scikit-beam here in this old channel for @dmgav to successfully deploy and test the software on the new TES server.

dmgav:
I think for 2019-C3.0 there is no need to update PyXRF version. I will install and update everything by hand.

mrakitin:
OK, maybe then this PR is not needed. In NSLS-II/playbooks#346, we can just have no extra packages for the analysis TES server. Does it make sense to everyone?

dmgav:
Well, I think that installing old version of PyXRF still helps to install the new version, because of dependencies.

mrakitin:
OK, let's stick to the original plan.

danielballan:
Sorry I came late to the party here. Yes, I agree is makes sense to use the old system for any `2019-*` environments.

mrakitin:
Thank you Dan!

PR #830, opened 2020-04-30
Bumped version to 0.1.9

stuwilkins:
Bumped centroids (pycentroids) to version 0.1.9. This version removes the dependence on `lmfit-c` as this is now included in the repo and build


mrakitin (approved):
Probably the version can be 0.2.0 with this relatively significant code change, but it's just a thought. Thanks!

